PR #133, opened 2025-04-02
test: add tests for Goto Pixel tab in Geo Coords Widget

Joooshe:
This MR does what the title says.

This only adds one tests, although it is a pretty important integration tests, but it creates the functionality needed to interact with the Goto Pixel tab. This will allow future tests to be created for this piece with much more ease.

I problem that I ran out of time to fix is that the tests to open up the Geo Coord dialog makes it so when you close the dialog, you can not interact with the main window. Currently, this has only caused an inconvenience when I am testing my tests by having the Gui show what it is doing. This does not affect the test implemented in this MR.

### DIFF
--- a/src/test_utils/test_model.py
+++ b/src/test_utils/test_model.py
@@ -84,7 +84,10 @@ class WiserTestModel:
         self.testing_widget = self.main_window._invisible_testing_widget
     
     def run(self):
-        QTimer.singleShot(10, self.app.quit)
+        def func():
+            self.app.processEvents()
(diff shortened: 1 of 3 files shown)


Joooshe:
As of now this is a low priority that hasn't been looked at in a while. Significant changes will likely be needed to how the test interacts with the test_model to make it a GUI test. This shouldn't be too hard to do, but will not work on it now, so closing. Here is an open issue for it: #533

PR #410, opened 2026-02-13
build: Fix Linux Build Errors

### LINKED ISSUE #411: Linux build not working due to opencv recursion error
## Describe the bug
The linux builds aren't working due to an opencv recursion error. This bug happened when I introduced the splash screen into WISER.

## To Reproduce
1. Make the _Build and Smoke Test WISER_ workflow run on commit b3a6d82f04d970d79e79cd04a8c007537d715a66 
2. See the cv2 recursion error occur when we smoke test WISER

## Expected behavior
The smoke tests should run without an error.

## Screenshots
_If applicable, add screenshots to help explain your problem._

## Desktop (please complete the following information):
Following platforms: Ubuntu amd and arm, Debian amd and arm, Fedora amd and arm.

## [Optional] Error Message
_The error message in the console if there is one._

## Additional context
_Add any other context about the problem here._

### LINKED ISSUE #412: On debian arm64, libsqlite3.so.0 not found error
## Describe the bug
We get the error that libgdal.so can't find libsqlite3.so.0 because its search path for libsqlite3.so.0 is in /opt/micromamba/envs/.../libsqlite3.so.0 which is on the build machine.

## To Reproduce
1. Run the _Build and Smoke Test WISER_ github action on commit 7a72409a8537d91a79e110caf089ef345cbe9959 
2. Wait for Debian arm64 to smoke test WISER then the error will happen

## Expected behavior
The smoke tests to succeed for debian arm64

## Screenshots
_If applicable, add screenshots to help explain your problem._

## Desktop (please complete the following information):
On the github runner in the docker file for debian arm64 (and so on other debian arm64 machines).

## [Optional] Error Message
_The error message in the console if there is one._

## Additional context
_Add any other context about the problem here._

Joooshe:
## What does this change do?
We are getting the opencv recursion error again but only on linux builds (Closes #411). To fix, I am trying to:
1. Make sure the .spec between `WISER-ubuntu.spec` and `WISER-macOS.spec` match perfectly. 
2. Compare the differences between the commits where it did work and where it stopped working. It seems that the difference was adding the splash screen.

We are also getting the libsqlite error (Closes #412). I think the solution I need for this is to search every `*.so`'s dependency list in their dynamic section and ensure none of the DT_NEEDED names are paths on the build machine. Also previously we overwrote DT_RPATH. Both overwriting it and using it in the first place is bad practice. Using DT_RPATH treats dependencies as transitive (which means if `A` depends on `B` and `B` depends on `C`, then when `C` is imported, `A` and `B`'s RPATH will still be used), using DT_RUNPATH means `A`'s RPATH will not be used when importing `C`, only `B`'s will). We will also need to make sure that no RPATH is hard coded to a path on the build machine. If it is, then we add $ORIGIN to that RPATH.

- [ ] Feature
- [x] Bug Fix
- [ ] Documentation Update
- [ ] Style
- [ ] Code Refactor
- [ ] Test
- [ ] Hot Fix
- [x] Build
- [x] CI
- [ ] Revert

## Why is this change needed?
This changed is needed so Linux builds work.

1. The open cv error was fixed by removing the splash screen code.
2. The libsqlite3 error was fixed by implementing a more general solution for removing built-time path's from a `.so`'s DT_NEEDED list.
    1. Previously we just patched libgdal.so's DT_NEEDED filename (patched from /opt/micromamba/.../libsqlite3.so.0 to libsqlite3.so.0 with symlink). Now the solution looks through all `*.so`s in the `_internal` folder in each of their dynamic sections in their ELF and ensures that there are no DT_NEEDED's with a build path to them and if their are it rewrites the build-time path to the correct soname (e.g. libsqlite3.so.0). This does assume there is a `.so` with a name that matches the build-time path, which most of the time there is. A future PR can update the code to look for a `.so` that matches the base name (without the .so.X).
    2. It also explicitly targets libsqlite3.so.* to make sure its symlinks exist. We also make sure the RPATH and RUNPATH has $ORIGIN appended to it if the path only has build-time paths. 

## Added tests?
- [ ] yes
- [X] no, because they aren’t needed (the test is the build step)
- [ ] no, because I need help

## Added to documentation?  
- [ ] yes
- [x] no documentation needed 

## Checklist
- [x] There is an issue associated with this pull request
- [x] Branch is up-to-date with main/master
- [x] All CI checks passed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ Secret.sh
 
 src/wiser/bandmath/temp_output/
 src/wiser/profiling/output/*
+
+dev_config_dir_windows/
--- a/WISER-macOS.spec
+++ b/WISER-macOS.spec
@@ -87,7 +87,7 @@ existing_hidden_imports = _hidden
 # pyinstaller_hooks/pyi_rth_cv2.py is meant to help fix this as well.
 cv2_binaries = collect_dynamic_libs(
     "cv2",
-    search_patterns=["cv2*.so", "cv2*.dylib", "python-*/cv2*.so", "python-*/cv2*.dylib"]
+    search_patterns=["cv2*.so", "cv2*.dylib", "python-*/cv2*.so", "python-*/cv2*.dylib"],
 )
 
 existing_binaries += cv2_binaries
--- a/WISER-ubuntu.spec
+++ b/WISER-ubuntu.spec
@@ -104,7 +104,7 @@ existing_hidden_imports = _hidden
 # OpenCV dynamic libs fix: collect cv2 shared objects (.so only on Linux)
 cv2_binaries = collect_dynamic_libs(
     "cv2",
-    search_patterns=["cv2*.so", "python-*/cv2*.so"],
+    search_patterns=["cv2*.so", "cv2*.dylib", "python-*/cv2*.so", "python-*/cv2*.dylib"],
 )
 existing_binaries += cv2_binaries
 
--- a/install-linux/multistage/Dockerfile
+++ b/install-linux/multistage/Dockerfile
@@ -83,8 +83,8 @@ RUN micromamba run -n "${ENV_NAME}" make -C /work build-linux
 
 # Bundle ELF dependency scrub/verify:
 # Conda absolute paths can leak into DT_NEEDED/RUNPATH (notably GDAL -> sqlite),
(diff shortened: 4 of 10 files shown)


Joooshe:
The OpenCV issue was resolved by #458 

Tee libsqlite error disappeared. I think the library or however the library was built must have fixed to not use the dev machines paths.

PR #506, opened 2026-04-23
feat: Adding full kmeans to WISER

### LINKED ISSUE #456: K-means algorithm
We should have a clustering algorithm that does basic K-means on data. We want to implement k-means and k-medoids with the option of difference distance functions for k-medoids. 

Select number "k", then do the rest.

More data analysis tools

### LINKED ISSUE #497: K Means Full Feature
We need the K-means backend to be built out such that we can get a y,x,1 size map that has integers for each k group. We need to be able to display a SpectrumPlotGeneric with the centroids for each of these groups. If possible, we should have a button that syncs the color of the spectrum plot spectra to the cmap colors in the main view.

### LINKED ISSUE #498: Build out k means backend
Should be pretty standard. We will get rid of all the data ignore pixels and then get rid of the bad bands so that the rest of the spectrum is squished in on itself. Then we will perform k means to get the group IDs and the centroids. We will map these group IDs back to a (y,x,1) array and the group IDs will be mapped to the centroids.

### LINKED ISSUE #499: K-means visualization
Once the K-means algorithm runs and we have the ID per pixel and we have the ID per centroid we need a way for the user to easily be able to map from ID to pixel. When the (y,x,1) dataset is created, a color map will be chosen for it. This is standard. The user should be able to go into the k-means dialog and click a button that pops up a widget. The widget should allow the user to click on k-means datasets that will then pop up the spectrum plot generic that will display the centroid spectra. The centroids will be named '1 - Centroid' so the user knows the int it corresponds to. If we can we will use the same color map for the color of these spectra as the color map that is being used in the main view for the dataset that corresponds to the clustered dataset.

### LINKED ISSUE #526: test_kmeans.py file stalls forever (12+ hours) during a test
## Describe the bug
This happens when running the second test in test_kmeans.py or if you run sklearn.cluster.KMeans.fit_predict before submitting the sklearn task. The test will just take forever and the function won't run (as evident by no prints to console).

## To Reproduce
Look at the bug description

## Expected behavior
The test to run and not stall forever

## Desktop (please complete the following information):
 - Linux Ubuntu Github runners (might also just be on all linux)

## [Optional] Error Message
No error message

Joooshe:
## What does this change do?
As the title says, this change adds the frontend, backend, and tests for the kmeans feature to WISER.

Closes #498 Closes #497 Closes #499 Closes #456 Closes #526

- [X] Feature
- [X] Bug Fix
- [ ] Documentation Update
- [ ] Style
- [ ] Code Refactor
- [ ] Test
- [ ] Hot Fix
- [ ] Build
- [ ] CI
- [ ] Revert

## Why is this change needed?
We need to add more data analysis tools to wiser. Kmeans is one.

Added KMeans as a semantic task and hooked it up so it can run from the GUI.

## Added tests?
- [X] yes
- [ ] no, because they aren’t needed
- [ ] no, because I need help

## Added to documentation?  
- [ ] yes
- [X] no documentation needed 

## Checklist
- [X] There is an issue associated with this pull request
- [X] Code compiles/builds without errors
- [X] No new lint/style issues introduced
- [X] Branch is up-to-date with main/master
- [X] All CI checks passed

## Oscar Feedback
- Have option for centroid spectra to be put into the spectrum plot  #527
- Try to make the spectra from the centroids match the color scheme in the raster view (maybe just make it all tab20 and say it in the name of the spectrum).  #528
- After k means write back in the no data value (for image pixels and for spectra). Fix is in this PR.


Joooshe:
Currently, the test called test_semantic_task_adds_labels_dataset_matching_sklearn_on_jpl is what is taking forever. It must be something in the function because the function does get submitted, it just never finishes.

Joooshe:
The issue seems to be a pthread_atfork() issue. On Linux the fork method for multiprocessing is used. ( I have tried ot make it use spawn with no luck). Using fork and multithreading on linux is [known to be hazardous](https://stackoverflow.com/questions/14407544/mixing-threads-fork-and-mutexes-what-should-i-watch-out-for). It seems to bite us here. Specifically, it seems that when fit_predict is called, OpenMP is used and GCC's OpenMP runtime is initialized. This is a [known issue ](https://thomasjpfan.github.io/parallelism-python-libraries-design/). 

One plausible reason behind this fit_predict issue is that it uses GCC's OpenMP which has a persistent internal thread pool that holds onto mutexes. It also registers handles to pthread_atfork. The prepare handle it registers needs the mutexes that the internal thread pool uses. However, the mutexes are only released in very short bursts and immediately pibkced back up and eventually they are fully released after a count as been reached. However, if GOMP_SPINCOUNT is set to INFINITE or INFINITY, then they are never full released. If set to infinite, its essentially a deadlock when pthread_atfork is called. However, thats only if GOMP_SPINCOUNT is set to infinity.

However, I don't know this for certain. Claude came up with this after I narrowed down where the solution could be and I no longer have time to fact check what it says to figure out the exact solution. What I do know is if I get rid of the fit_predict call in the test_kmeans tests, that i don't have this issue, the issue only happens after the fit_predict call when the next fork() call is made (by submitting to the ProcessPoolExecutor), the function that is supposed to run in the subprocess does not run and this only happens on linux where fork is used.

**The Solution** is to use forkserver because the forkserver forks from a "fresh" server process. This means that the forked process gets created from a fresh process and so when it is created it doesn't have to wait for an mutexes that would be set up during a call to fit_predict so there will be no deadlocks.

Addresses #526

Issue #626, opened 2026-06-23
[Project File] Save flow + dependency-aware dialog


Joooshe:
## Is your feature request related to a problem? Please describe.

Even with every item persister built, the user needs to **understand and control** what gets saved — especially which in-memory datasets (and in-memory libraries) to include, since those are the only roots that require a decision and everything else cascades from them. Without a clear UI, users would unknowingly drop dataset-backed spectra, ROI-average spectra, or runs when they skip a dataset, and would not understand why something "didn't save."

## Describe the solution you'd like

A **dependency-aware Save dialog** plus the top-level save flow, driven by the resolver (#617).

**Mental model presented to the user:** "You're really only deciding which RAM-backed datasets and in-memory libraries to include. Everything else follows automatically."

**Dialog behavior:**
1. List the **savable roots** (RAM-backed datasets, in-memory libraries) with checkboxes. File-backed datasets/libraries are free and need no decision (shown as auto-included).
2. When a root is unchecked, **live-cascade** the consequence: show, per dependent item, whether it will be **FAITHFUL**, **SNAPSHOT** (with a one-line "will be saved as a frozen copy" note), or **DROP** (with a warning).
3. For DROP-eligible items, offer the **snapshot escape hatch** ("freeze this spectrum so it survives without its dataset") where snapshotting is possible.
4. Offer **"promote to file"** for a RAM dataset (write to an external path and reference it) as an optional alternative to a bundle sidecar.
5. Summarize warnings (what will be dropped) before the user confirms.

**Save flow:**
- Orchestrate writing the bundle: manifest (pyrep) + dataset sidecars + array sidecars, in dependency order.
- Choose bundle directory vs. zipped `.wiserproj`.
- Wire up menu actions (Save Project / Save Project As) and remember the last project path.

## Describe alternatives you've considered

- **Save everything with no dialog.** Rejected: users can't control bundle size or understand drops; large RAM cubes would always be written.
- **Present the entire dependency tree as individually-toggleable nodes.** Rejected: overwhelming; the root-checkbox + automatic-cascade model matches the DAG (datasets are the only real decision points).
- **Block saving until all dependencies are saveable.** Rejected: too rigid; snapshot/drop with warnings is friendlier.

## Additional context

- Depends on the resolver/policy #617, datasets #618, and all item persisters #619–#625 (it orchestrates their save side).
- Pairs with load/restore #627.

Issue #619, opened 2026-06-23
[Project File] Region-of-interest persistence


Joooshe:
## Is your feature request related to a problem? Please describe.

Regions of interest are `[SOURCE]` state (`ApplicationState._regions_of_interest`) and are lost on close. They must be restored **faithfully** — as their original geometry (polygon corner points, rectangle corners, individual pixels). The existing ROI-level serialization helpers cannot do this today:

- `roi_to_pyrep` / `roi_from_pyrep` in [roi.py](https://github.com/Ehlmann-research-group/WISER/blob/094187c1/src/wiser/raster/roi.py) are **stale**: they call `roi.get_selection()` (singular — no such method), assume a single selection (`TODO(donnie): Composite/multi-selection ROIs`), and `from_pyrep` calls a `RegionOfInterest(...)` constructor signature that no longer exists.
- `RegionOfInterest` actually holds a **list** of selections (`_selections`).

The good news: the per-selection layer is already solid — every `Selection` subclass has a working `to_pyrep`/`from_pyrep` ([selection.py](https://github.com/Ehlmann-research-group/WISER/blob/094187c1/src/wiser/raster/selection.py)).

## Describe the solution you'd like

Persist each ROI as its full identity plus a list of faithfully-serialized selections, and rewrite the broken ROI-level helpers.

**Save** (`roi_to_pyrep`, rewritten): `{id, name, color, description, metadata, selections: [sel.to_pyrep() for sel in roi.get_selections()]}`.

Per-selection pyreps already capture geometry exactly:
- `PolygonSelection` → list of corner **points** (as requested — points, not rasterized pixels).
- `RectangleSelection` → two corner points.
- `SinglePixelSelection` / `MultiPixelSelection` → the point(s).

**Load** (`roi_from_pyrep`, rewritten): rebuild each selection via `selection_from_pyrep`, append all to a fresh `RegionOfInterest`, restore id/name/color/description/metadata.

**No dependency on datasets.** Selections are pure pixel coordinates with no bound dataset, so ROIs save standalone and, on restore, land on whatever dataset they're applied to (off-grid is acceptable — matches current behavior).

**Notes / cleanups in scope:**
- Fix the multi-selection handling (the core of this issue).
- WISER does **not** support predicate ROIs, so `PredicateSelection` need not be exercised — but if its pyrep is kept, fix the existing `from_pyrep` bug (`points=` should be `predicate=`) or explicitly exclude it.
[...]

Issue #618, opened 2026-06-23
[Project File] Dataset persistence (sidecars + id preservation)


Joooshe:
## Is your feature request related to a problem? Please describe.

Nearly every other saveable item in WISER references a dataset by numeric id — spectra, stretches, run records, and ROI-average spectra all point back to one. This makes datasets the foundation everything else is built on, and the most important item to get right when saving. Two specific problems make it harder than it sounds:

1. **In-memory datasets have no file to point at.** A `RasterDataSet` whose `_impl` is an in-memory array (tool outputs, some loads) has no path; its pixels must be captured or the dataset is lost (see [app-state.md](https://github.com/Ehlmann-research-group/WISER/blob/094187c1/doc/sphinx-general-wiser-docs/source/developer-content/app-state.md), Datasets section).
2. **Dataset ids hold everything together.** Stretch keys, run-record `input_dataset_id`/`output_dataset_id`, and spectrum sources all reference a dataset by its integer id. If those ids are reassigned to different datasets on load, every reference silently points at the wrong thing.

## Describe the solution you'd like

This issue covers both directions — **saving** datasets into the project file and **loading** them back out. Both directions are described below under their own headings. Datasets must be handled first in both directions: on save, they are written before anything that references them; on load, they are restored before anything that references them, so that the restored session has datasets to link against when spectra, stretches, and run records are rebuilt.

**Save:**

- **File-backed dataset** → store its filepath(s) (`_impl.get_filepaths()`) plus identity/metadata, *by reference*.
- **In-memory dataset** → write a **raster sidecar** into the bundle (`datasets/ds_<id>.*` via GDAL/ENVI/GeoTIFF) and reference it by relative key. This is the default ("sidecar") behavior — the user does not have to manually save RAM datasets.
- Optionally expose a **"promote to file"** action so a user can write a RAM dataset to a chosen external path and have the project reference that instead (escape hatch, not required).
- Capture the dataset's `[SOURCE]` metadata: `_name`, `_description`, `_band_unit`, `SpectralMetadata` (band_info, bad_bands, default_display_bands, data_ignore_value, has_wavelengths), and `SpatialMetadata` (geo_transform, wkt_spatial_reference). This can be edited by the user at runtime so it may not be apart of the `_impl` type in the `RasterDataSet`. `_cached_band_stats` is `[DERIVED]` — do **not** save.
- Record the dataset's **numeric id** in the manifest.

**Load (fresh session only — no merge):**

[...]

Issue #615, opened 2026-06-23
[EPIC] Save / Open WISER Project File


Joooshe:
## Is your feature request related to a problem? Please describe.

Today a WISER session is entirely in-RAM. When the user closes WISER, **everything they configured is lost**: loaded datasets, the spectra they picked and collected, their regions of interest, contrast stretches, PCA/MNF/linear-unmixing/K-Means run histories, user-created CRSs, and saved band-math expressions. There is no way to:

- close WISER and resume the same analysis later,
- reproduce a result that depended on transient, RAM-only artifacts (e.g. an unseeded K-Means run, an in-memory dataset that was never written to disk).

## Describe the solution you'd like

A **"Save to Project File" / "Open Project File"** feature that captures and restores a WISER session.

The project is saved as a **bundle** (a directory, optionally zipped): a small, human-readable, version-tolerant **manifest** (plain-dict "pyrep" → JSON) plus **raster sidecar files** for any in-memory datasets. The manifest references file-backed datasets by path and embeds the small state items directly.

The design is governed by one rule, derived from the fact that the state forms a **DAG rooted at datasets**:

> **Reconstruct faithfully when a dependency is present; snapshot it when the dependency is cut; drop it only if the user explicitly declines the snapshot — never write a dangling reference.**

Scope of what is saved (each its own sub-issue):

- **Datasets** — file-backed by reference, in-memory become sidecars; IDs preserved on load (see #618).
- **Regions of interest** — faithful geometry, per selection (see #619).
- **Active + collected spectra** — reference / snapshot / drop per the rule (see #620).
- **Spectral libraries** — path reference or RAM snapshot (see #621).
- **Contrast stretches** — per `(dataset, band)` (see #622).
- **Run histories** (PCA / MNF / linear unmixing / K-Means) — records always saved, datasets opt-in, missing → "closed run" (see #623).
- **User-created CRSs** — WKT + construction state (see #624).
- **Band-math saved expressions** (see #625).

Underpinned by foundation work: the **bundle format + pyrep conventions** (#616), the **dependency resolver + save policy engine** (#617), the **save flow + dependency-aware dialog** (#626), the **load/restore orchestration** (#627), and the **versioning & migration subsystem** (#628) that guarantees older project files keep opening.

[...]

PR #664, opened 2026-07-02
feat(project): save-to-project bundle foundation + ROI persister (#616, #619)

### LINKED ISSUE #616: [Project File] Bundle format + pyrep serialization conventions

The "Save to Project File" epic (#615) needs a concrete, durable on-disk format before any individual state item can be persisted. WISER has two existing serialization styles and neither is currently suited to long-lived project files:

- **pyrep** (`to_pyrep`/`from_pyrep`) — plain Python dicts, used today by `Selection` types ([selection.py](https://github.com/Ehlmann-research-group/WISER/blob/094187c1/src/wiser/raster/selection.py)). JSON-friendly and portable, but not used consistently across state items, and many items (e.g. stretches) have no pyrep at all.
- **`Serializable` / `get_serialized_form()`** — a pickle-style mechanism used for **cross-process** transfer (e.g. shipping datasets to worker processes, see `Selection.__getstate__`). It is brittle across code versions and unsafe for files received from others.

Without a single agreed format, every sub-issue would invent its own, producing an inconsistent and unmigratable project file.

Define and implement the **project bundle format** and a **shared pyrep serialization convention** that all other sub-issues build on.

**Bundle layout** — a project is a directory (optionally zipped to a single `.wiserproj` file):

```
my_project.wiserproj/            (dir or zip)
  manifest.json                  # the pyrep of the whole session
  datasets/                      # raster sidecars for in-memory datasets
    ds_7.img / ds_7.hdr ...
  arrays/                        # optional .npy sidecars for large arrays
  format_version                 # integer, for migration
```

**Manifest = pyrep (plain dicts) → JSON.** Decision (see epic discussion): use **pyrep for the manifest**, *not* the `Serializable`/pickle path. Rationale: portability, human-readability, version tolerance, and safety for shared files. The `Serializable` path stays where it belongs (cross-process), and **bulk raster bytes never go in the manifest** — datasets write themselves as native raster sidecars (GDAL/ENVI/GeoTIFF).

**Deliverables:**

1. A `ProjectBundle` abstraction that can create/open a bundle (dir or zip), read/write `manifest.json`, and add/fetch sidecar files (rasters, `.npy` arrays) by a stable relative key.
2. A top-level manifest schema with `format_version`, plus a top-level section per state item (filled in by sub-issues 03–10).
3. A documented **pyrep convention**: every persistable type exposes `to_pyrep() -> dict` and a classmethod/`from_pyrep(data)`; each dict carries a `"type"` tag; large numpy arrays are written to the `arrays/` sidecar area and referenced by key rather than inlined.
4. A `format_version` + the migration seam (the migrate-up mechanism itself; the migration *policy* and test suite are owned by #628).
5. Round-trip unit tests for the empty/skeleton manifest and the sidecar mechanism.

**Backwards-compatibility strategy (migrate-up on load).** The format is built to evolve from day one, so this is a first-class concern rather than an afterthought:

- **Backward compatibility is guaranteed; forward compatibility is not.** A *newer* WISER must always open *older* project files. An *older* WISER is **not** expected to open files from a newer one — instead, when the loader sees a `format_version` greater than it understands, it **refuses cleanly** ("this project was made with a newer WISER — please upgrade") rather than crashing or silently dropping data.
- **One current-schema loader + a migration chain.** The loader only ever understands the **current** schema. On load, if the file is older, a chain of small *pure* functions (`v1→v2→v3→…→current`) transforms the manifest *pyrep dict* up to the current shape, then the normal load proceeds. All version-handling is isolated in this one migration layer; sub-issues 03–12 only ever deal with the current schema. (Mirrors DB migrations / QGIS.)
- **Lenient additive parsing avoids most migrations.** For non-breaking changes (a new *optional* field), don't bump `format_version` or write a migration — `from_pyrep` ignores unknown keys and supplies defaults for missing ones. Bump the version and write a migration only for **breaking** changes (renamed / removed / restructured fields).
- **Single global `format_version`** for the whole bundle (migrations operate on the whole manifest dict), not per-section versions. Per-section versioning can be added later if one item churns heavily.

The migration-chain harness, the refuse-too-new behavior, and the regression suite that *enforces* "old files still load" are specified in #628.

A portable, human-readable, version-tolerant project format is what makes sessions **shareable across machines and OSes** (WISER runs on macOS/Windows/Linux) and **reproducible across WISER releases**. Avoiding pickle also keeps shared project files safe to open.

- **A single binary/pickled file.** Rejected: not version-tolerant, not safe to share, opaque to debugging.
- **One flat JSON with base64-embedded rasters.** Rejected: huge, slow, defeats native raster tooling.
- **Reusing `Serializable` for the whole manifest.** Rejected for storage (see above); kept for cross-process only.
- **YAML/TOML instead of JSON.** JSON chosen for ubiquity and tooling; the pyrep dicts are format-agnostic so this can change later behind the `ProjectBundle` seam.

## Additional context

- Foundation issue: blocks **all** of 02–13.
- The versioning/migration *policy* and its enforcing test suite live in #628; this issue only provides the `format_version` field and the migration seam.
- Existing pyrep reference implementation to mirror: the `Selection` subclasses in [selection.py](https://github.com/Ehlmann-research-group/WISER/blob/094187c1/src/wiser/raster/selection.py).
- Bundle-as-directory-or-zip mirrors how QGIS projects and many geospatial tools ship sidecar data alongside a manifest.

mmaclay:
## What does this change do?
Introduces the foundation for saving and restoring a WISER session as a portable `.wiserproj` bundle, plus the first per-item persister (ROIs) that proves the serialization convention end-to-end. Foundation layer of the Save/Open Project epic (#615).

- [x] Feature
- [x] Bug Fix
- [ ] Documentation Update
- [ ] Style
- [x] Code Refactor
- [x] Test
- [ ] Hot Fix
- [ ] Build
- [ ] CI
- [ ] Chore
- [ ] Revert

## Why is this change needed?
A WISER session is entirely in-RAM today — closing the app loses loaded datasets, ROIs, spectra, stretches, run histories, etc. This PR lands the on-disk bundle format and the shared serialization convention that every other persister in the epic builds on.

Closes #616 (bundle format + pyrep conventions) and #619 (ROI persister). Part of epic #615.

New `src/wiser/project/` package:
- **`ProjectBundle`** — a bundle is a directory (a `.wiserproj` is that directory zipped): `manifest.json` (a plain-dict "pyrep" → JSON) plus `arrays/`
(`.npy`) and `datasets/` (raster) sidecars. Bulk data never ent
- **`pyrep`** — the shared convention: every persistable type eep()`; each dict carries a `"type"` tag; a dispatch registry
reconstructs by tag; large arrays are referenced by key. `from_ding an optional field needs no migration.
- **`migrate`** — `format_version`, a migrate-up chain, and a c
- **`persisters/rois.py`** (#619) — the canonical persister thep`/`roi_from_pyrep` rewritten to round-trip multi-selection ROIs
faithfully.

Also in this PR:
- Fixes two latent `selection.py` `from_pyrep` bugs surfaced by the round-trip: `SinglePixelSelection` never rebuilt its `QPoint` (restored single-pixel
ROIs would crash), and `PredicateSelection` used the wrong cons
- Removes the disabled + broken legacy `.wiser` save/load path en_file` routing — the `.wiserproj` bundle supersedes it.
- Adds an optional `ds_id` parameter to `ApplicationState.add_dataset` as groundwork for id-preserving dataset restore (consumed by the stacked #618 PR).

## Added tests?
- [x] yes

`src/tests/test_project_bundle.py` — pyrep round-trips, registry dispatch, unknown-type handling, bundle manifest + array sidecar, zip round-trip, migration (no-op + refuse-too-new), and the ROI persister round-trip. Focused suite green: **29 passed** (`test_project_bundle` + `test_roi_storage` +
`test_selection_pickle`).

## Added to documentation?
- [x] no documentation needed

User-facing Save/Open UI and docs land with the orchestration issues #626/#627.

## Checklist
- [x] There is an issue associated with this pull request (#616
- [ ] Branch is up-to-date with main/master — based on the PySide6 merge (`8d8c53f1`); **needs a rebase onto current `main` before merge**

- Rebase the stacked #618 PR (`feat/proj-618-datasets`) onto `main` once this merges.
</details>

### DIFF
--- a/doc/sphinx-general-wiser-docs/source/developer-content/code-documentation.md
+++ b/doc/sphinx-general-wiser-docs/source/developer-content/code-documentation.md
@@ -22,4 +22,5 @@ georeferencer-internals.md
 crs-creator-internals.md
 plugin-system.md
 mosaic-internals.md
+save-to-project.md
 ```
--- a/doc/sphinx-general-wiser-docs/source/developer-content/save-to-project.md
+++ b/doc/sphinx-general-wiser-docs/source/developer-content/save-to-project.md
@@ -1,0 +1,223 @@
+# Save to Project File
+
+WISER can persist an entire working session — the loaded datasets, regions of
+interest, collected spectra, contrast stretches, spectral libraries, analysis run
+histories, user-created coordinate systems, and saved band-math expressions — to a
+single portable **project file** and later reopen it into a fresh session. This
+page documents the on-disk format, the serialization convention, the dependency
+model that decides what is saved faithfully versus snapshotted, and the save/load
+orchestration.
+
+The feature lives in `src/wiser/project/`. It is deliberately layered so that each
+piece of session state is serialized by an isolated *persister*, and a single
+orchestrator drives them in dependency order.
+
+---
+
+## Overview
+
+A project file is a **bundle**: a `manifest.json` describing the session, plus two
+sidecar directories for the binary data that does not belong inline in JSON.
+
+| Path | Contents |
+|------|----------|
+| `manifest.json` | The whole session as type-tagged JSON — one section per state kind. |
+| `arrays/` | NumPy `.npy` sidecars for arrays too large for the manifest (e.g. in-memory dataset pixels, endmember snapshots). |
+| `datasets/` | ENVI-format raster sidecars for in-memory (RAM-backed) datasets that have no source file on disk. |
+
+A bundle is written either as a **directory** or, when the destination ends in
+`.wiserproj`, as a single **zip** of that directory. File-backed datasets are
+stored *by reference* (their path), so a project that loads only on-disk rasters is
+tiny; only RAM-backed datasets and their derived arrays add bulk.
+
+```{mermaid}
+flowchart TB
+    subgraph UI["GUI layer"]
+        Menu["File menu<br/>Open / Save / Save As"]
+        Dialog["SaveProjectDialog<br/>choose RAM datasets, preview cascade"]
+    end
+    subgraph Orch["Orchestration — orchestrate.py"]
+        Save["save_project()"]
+        Load["load_project()"]
+    end
+    subgraph Pers["Persisters — persisters/*.py"]
+        P["datasets · rois · spectra · stretches<br/>libraries · runs · crs · bandmath"]
+    end
+    subgraph Core["Foundation"]
+        Bundle["ProjectBundle<br/>manifest + sidecars, dir/zip"]
+        Resolver["DependencyResolver<br/>FAITHFUL / SNAPSHOT / DROP"]
+        Migrate["migrate.py<br/>format_version, migrate-up"]
+    end
+    Menu --> Save & Load
+    Dialog --> Resolver
+    Save --> P
+    Load --> P
+    P --> Bundle
+    Save --> Resolver
+    Load --> Migrate
+    Bundle --> Migrate
+```
+
+---
+
+## The pyrep convention
+
+Every serializable object is converted to a **pyrep** ("Python representation"): a
+JSON-safe `dict` carrying a `"type"` tag that names how to rebuild it. A central
+registry maps each tag to a reconstruction function.
+
+- `register_pyrep(tag, from_fn)` registers a reconstructor; `from_pyrep(data)`
+  dispatches on `data["type"]`.
+- Large arrays are not inlined. A persister calls the bundle to stash the array and
+  writes an **array reference** in its place — `array_ref(key)` produces a small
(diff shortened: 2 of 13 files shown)


Joooshe (changes requested):
Looks good. I would add a page to the docs as you build out this feature and keep it updated. Mermaid diagrams would be helpful as well. This would probably go in the Developer Documentation section, but you may want to make a sub folder. 

I really like your solution to be defensive against malicious inputs!

Joooshe (approved):
Looks good. It will be important to keep save-to-project.md updated with each new PR.

PR #665, opened 2026-07-02
feat(project): dataset persister with ENVI sidecars (#618)

### LINKED ISSUE #618: [Project File] Dataset persistence (sidecars + id preservation)

mmaclay:
## What does this change do?
Adds the dataset persister for the Save/Open Project feature: saves and restores `RasterDataSet`s in a project bundle, with datasets as the id-preserving backbone every other item references.

- [x] Feature
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] Style
- [ ] Code Refactor
- [x] Test
- [ ] Hot Fix
- [ ] Build
- [ ] CI
- [ ] Chore
- [ ] Revert

## Why is this change needed?
Datasets are the roots of the project's dependency DAG — restoring them with their original ids intact is what keeps downstream references (spectra,
stretches, run records) valid across a save/load. Closes #618.

**Stacked on #616/#619** (`feat/proj-file`). This PR is based oows only the #618 changes — please review/merge #616/#619 first, after which this rebases onto `main`.

New `src/wiser/project/persisters/datasets.py`:
- File-backed datasets (`get_filepaths()` non-empty) are saved **by reference** — only the path is recorded; bulk bytes never enter the bundle.
- In-memory datasets (NumPy-backed, e.g. a band-math result) arNVI sidecars (`datasets/ds_<id>.img` + `.hdr`) via the existing`RasterDataLoader.save_dataset_as`.
- Restore re-opens each dataset and re-registers it with its orte.add_dataset(ds_id=...)`.
- A referenced file that has since moved (or a missing sidecar) is reported as **dropped**, not fatal — the load warns and continues rather than leaving a
dangling reference.

## Added tests?
- [x] yes

`src/tests/test_project_datasets.py` — the sidecar and by-reference round-trips, id preservation, the zip round-trip, and the missing-reference drop. **4
passed.**

## Added to documentation?
- [x] no documentation needed

## Checklist
- [x] There is an issue associated with this pull request (#618
- [x] No new lint/style issues introduced (ruff + ruff-format c
- [ ] Branch is up-to-date with main/master — stacked on `feat/proj-file`; rebase onto `main` after that merges

## [optional] Are there any post-merge tasks we need to perform
- After #616/#619 merges, rebase this branch onto `main` (GitHub retargets the base automatically), then this PR closes #618.

### DIFF
--- a/src/tests/test_project_datasets.py
+++ b/src/tests/test_project_datasets.py
@@ -10,16 +10,21 @@ a malformed id.
 import os
 
 import numpy as np
+from astropy import units as u
 
 import tests.context  # noqa: F401
+
+from osgeo import osr
 
 from wiser.project import ProjectBundle, unzip_bundle, zip_bundle
 from wiser.project.persisters.datasets import (
     STORAGE_REFERENCE,
     STORAGE_SIDECAR,
+    dataset_to_pyrep,
     load_datasets,
     save_datasets,
 )
+from wiser.raster.dataset import SpatialMetadata
 from wiser.raster.loader import RasterDataLoader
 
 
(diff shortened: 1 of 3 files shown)


src/wiser/project/persisters/datasets.py
```diff
@@ -0,0 +1,165 @@
...
+        if dataset is None:
+            dropped.append(ds_id)
+            continue
+        app_state.add_dataset(dataset, ds_id=ds_id)
+
+    return dropped
+
+
+def _load_dataset(
+    entry: Dict[str, Any], bundle: ProjectBundle, loader: "RasterDataLoader", cache
```
Joooshe: NIT: It would be nice to type the cache here as well

src/wiser/project/persisters/datasets.py
```diff
@@ -0,0 +1,165 @@
...
+    entry: Dict[str, Any] = {
+        "type": DATASET_PYREP_TYPE,
+        "id": ds_id,
+        "name": dataset.get_name(),
+        "description": dataset.get_description(),
+    }
+
+    filepaths = dataset.get_filepaths()
+    if filepaths:
+        entry["storage"] = STORAGE_REFERENCE
```
Joooshe, line 79: Correct me if I am wrong, but I don't think this would properly be able to reopen a netcdf dataset (or any dataset that is opened from a subdataset). You may need another optional metadata value in the entry on which subdataset the user had selected.

src/wiser/project/persisters/datasets.py
```diff
@@ -0,0 +1,165 @@
...
+    datasets_root = (bundle.root / ProjectBundle.DATASETS_DIR).resolve()
+    target = (bundle.root / rel).resolve()
+    try:
+        target.relative_to(datasets_root)
+    except ValueError:
+        return None
+    return str(target)
+
+
+def _apply_source_metadata(dataset: "RasterDataSet", entry: Dict[str, Any]) -> None:
```
Joooshe, line 189: I may have missed something in the PR, but is the idea of keeping metadata across sessions of WISER that the metadata is saved into the sidecar or gdal dataset? If so, i don't think the sidecars have their datasets saved. It seems like they just have hyper spectral array data saved. Also, for the gdal datasets, the user can alter a RasterDataSet's in the GUI. If we want to fully recreate the metadata that they had altered, we would need to snapshot the RasterDataSet's metadata state and save that in either the manifest or a metadata specific sidecar (metadata can have arrays, for example bad band list (bbl) can be a 425 length array).

Joooshe (changes requested):

src/wiser/project/persisters/datasets.py
mmaclay: You're totally right! a NetCDF subdataset was being dropped on load. Fixed this in the latest push: the manifest now records the base .nc path plus the subdataset_name, and load reopens via load_from_file(base, subdataset_name=…, interactive=False) so the same subdataset is restored rather than re-running the auto-pick heuristic. Note this is NetCDF-only.

src/wiser/project/persisters/datasets.py
mmaclay: Done, cache is annotated now

src/wiser/project/persisters/datasets.py
mmaclay: Good catch, you're right on both counts. latest push adds a JSON metadata snapshot to each manifest entry: data-ignore, bad bands (array-valued, e.g. the 425-length BBL you mention), wavelengths + units, band descriptions, display bands, and georeferenced CRS/geo-transform. It's reapplied on load via the safe per-field setters in a fixed order

src/tests/test_project_datasets.py
```diff
@@ -0,0 +1,347 @@
...
+        "datasets": [
+            {"type": "RasterDataSet", "name": "no-id", "storage": STORAGE_REFERENCE, "path": "x"},
+            {"type": "RasterDataSet", "id": "3", "name": "str-id", "storage": STORAGE_REFERENCE, "path": "x"},
+        ]
+    }
+    bundle = ProjectBundle.create(tmp_path / "proj")
+    dst = _FakeAppState()
+    # No integer id to preserve or report: skipped entirely, not in dropped.
+    assert load_datasets(manifest, dst, bundle) == []
+    assert dst.get_datasets() == []
```
Joooshe, line 411: Last thing. And sorry I should've said this on my last round of comments. I think we should test that netcdf saving and loading works since it uses a slightly different load mechanic. The netcdf file in `src\test_utils\test_datasets\netcdf.nc` it has a subdataset named `reflectance` and a subdataset named `dummy`.

Joooshe (dismissed):

Issue #622, opened 2026-06-23
[Project File] Contrast-stretch persistence


Joooshe:
## Is your feature request related to a problem? Please describe.

The current contrast stretch per band (`ApplicationState._stretches`, keyed by `(dataset_id, band_index)`) is `[SOURCE]` state lost on close. Without it, a restored session looks visually different (every band falls back to a default stretch), which is jarring and hides the analyst's chosen visualization. Two complications:

- Stretches are **keyed by dataset id**, so each is a leaf hanging off a dataset — saving them joins the dataset cascade (a stretch whose dataset isn't saved is simply dropped).
- Stretches have **no `to_pyrep` today** ([stretch.py](https://github.com/Ehlmann-research-group/WISER/blob/094187c1/src/wiser/raster/stretch.py)), and some types come in **numba jitclass** variants.

## Describe the solution you'd like

Add pyrep serialization to the stretch type hierarchy and persist `_stretches` keyed by `(dataset_id, band_index)`.

**Per-type serialization** (small closed set):
- `StretchLinear` → `lower`, `upper` (slope/offset derive). Trivial.
- `StretchSquareRoot`, `StretchLog2` → type tag only. Trivial.
- `StretchHistEqualize` → `_cdf` and `_histo_edges` (numpy arrays). **Snapshot the two arrays** (small; inline as lists or as `.npy` sidecars). Recomputing from the dataset histogram is also possible but snapshotting is simpler and robust.
- `StretchDecorrelation` → its derived state, same array-snapshot approach.
- `StretchComposite` → **must be handled.** A committed stretch is frequently a `StretchComposite` wrapping a conditioner **plus** a base stretch (e.g. SQRT conditioner + linear). It exposes the two halves via `first()` / `second()` (see how `StretchBuilder._load_individual_stretch` unpacks it, [stretch_builder.py:323-325](https://github.com/Ehlmann-research-group/WISER/blob/094187c1/src/wiser/gui/stretch_builder.py#L323-L325)). Serialize it by recursively serializing both halves; rebuild it from both on load. Omitting this would silently drop the conditioner on any conditioned stretch.

**Numba variants:** serialize **parameters, not the jitclass object**. `to_pyrep` pulls plain fields (name/scalars/arrays); `from_pyrep` calls the normal constructor, which can build either the numba or non-numba variant. The jitclass is reconstructed fresh at load — identical behavior.

[...]

Issue #621, opened 2026-06-23
[Project File] Spectral library persistence


Joooshe:
## Is your feature request related to a problem? Please describe.

Spectral libraries (`ApplicationState._spectral_libraries`) are `[SOURCE]` state lost on close, and were not initially considered in the save design. They behave "like datasets" (file-backed vs. RAM-backed) but with a twist: a `ListSpectralLibrary` is just a `List[Spectrum]` ([spectral_library.py](https://github.com/Ehlmann-research-group/WISER/blob/094187c1/src/wiser/raster/spectral_library.py)), so a single in-memory library can be **heterogeneous** — mixing `NumPyArraySpectrum` (self-contained) with `RasterDataSetSpectrum` members (dataset-backed). A naive whole-library snapshot would mishandle the dataset-backed members.

## Describe the solution you'd like

Persist libraries by reusing the spectrum logic from #620 per member — there is no special library-level dependency logic needed.

**Save:**
- **File-backed library** (`_path` set) → reference the path; store `_name`, `_description`, `_path`. Done.
- **RAM-backed library** (`_path is None`) → save `{name, description, path: null, spectra: [...]}`, where **each member spectrum is written to the manifest as a pyrep entry, following the same FAITHFUL / SNAPSHOT / DROP rule** as active/collected spectra in #620. Heterogeneity resolves itself because the decision was always per-spectrum.

**Load:** reconstruct file-backed libraries from their path. For RAM libraries, read each member spectrum's pyrep entry from the manifest and reconstruct it — as a live object if its dataset was restored (FAITHFUL), or as a `NumPyArraySpectrum` from the stored values if not (SNAPSHOT). This is the same pyrep-based reconstruction used for active/collected spectra in #620; `Serializable`/`SerializedForm` is not used here.

This keeps libraries conceptually simple: a library is just a named, optionally-path-backed list of spectra, and each member spectrum is saved and loaded by the same rules defined in #620 — no separate library-level logic needed.

## Describe alternatives you've considered

- **Treat every library as a monolithic snapshot.** Rejected: mishandles heterogeneous RAM libraries and dataset-backed members.
- **Only support file-backed libraries.** Rejected: RAM/curated libraries are common and would be silently lost.
- **Invent library-specific dependency rules.** Unnecessary — per-member reuse of #620 covers it.

## Additional context

- Depends on #616, spectra #620 (reused per member), and transitively datasets #618 for dataset-backed members.
[...]

Issue #620, opened 2026-06-23
[Project File] Active + collected spectra persistence


Joooshe:
## Is your feature request related to a problem? Please describe.

The **active spectrum** (`_active_spectrum`) and **collected spectra** (`_collected_spectra`) are `[SOURCE]` state lost on close. They are the trickiest spectra to persist because a `Spectrum` is polymorphic ([spectrum.py](https://github.com/Ehlmann-research-group/WISER/blob/094187c1/src/wiser/raster/spectrum.py)):

- `NumPyArraySpectrum` — self-contained values; no dependency.
- `RasterDataSetSpectrum` / `SpectrumAtPoint` — hold a **live dataset reference** (`self._dataset`) → depend on a dataset.
- `ROIAverageSpectrum` — holds a dataset **and** an ROI (`self._roi`) → **double dependency**.

If a backing dataset is not saved, a naive "reference the dataset" approach would write a dangling reference.

## Describe the solution you'd like

Persist active + collected spectra using the epic's governing rule, driven by the dependency resolver (#617):

1. **FAITHFUL** — if a spectrum's backing dataset (and ROI, for `ROIAverageSpectrum`) is being saved, re-instantiate the *real* live object on load (e.g. rebuild the `ROIAverageSpectrum` against the restored dataset + ROI; it recomputes the same average and stays live).
2. **SNAPSHOT** — if a dependency is cut, freeze the spectrum into a `NumPyArraySpectrum` capturing values + wavelengths (with units) + bad-band mask + `_name` + `_color` + `_source_name`. Data is preserved; only *liveness* is lost.
3. **DROP** — only if the snapshot is declined; record a warning.

`NumPyArraySpectrum` instances are always FAITHFUL (already self-contained). `_icon` is `[DERIVED]` — never saved.

**Manifest shape:** an ordered list for collected spectra and a reference to which one (if any) is active; each entry is a pyrep with a `kind` (`numpy` / `raster-backed` / `roi-average`) and either a dataset/ROI reference (FAITHFUL) or inline/`.npy` values (SNAPSHOT).

## Describe alternatives you've considered

- **Always snapshot every spectrum.** Rejected: loses liveness even when the dataset is present and round-trippable.
- **Always reference the dataset.** Rejected: writes dangling references when the dataset isn't saved.
- **Drop dataset-backed spectra whose dataset isn't saved.** Rejected as default: the snapshot fallback preserves the data; dropping is last resort.

[...]

PR #670, opened 2026-07-07
 feat(project): dependency resolver + spectra persister (#617, #620)

### LINKED ISSUE #617: [Project File] Dependency resolver + save-policy engine

WISER's saveable state forms a **dependency DAG rooted at datasets**: dataset-backed spectra depend on a dataset; ROI-average spectra depend on a dataset **and** an ROI; run records (e.g. `MNFRunRecord`, `LinearUnmixingRunRecord`) reference input/output datasets; stretches are keyed by `(dataset_id, band)`. When the user chooses *not* to save a dataset (e.g. a large RAM-only cube), everything that transitively depends on it becomes unsaveable. Without a single component that understands these edges, each sub-issue would re-implement "is this saveable?" logic ad hoc, leading to inconsistent behavior and writing a **dangling reference** into the project file.

A **dependency resolver + save-policy engine** that every item persister (what is responsibile for writing state to the project file for a specific part of the state) and the save project file dialog use.

**Responsibilities:**

1. **Build the dependency graph** for the current session: nodes = saveable items (datasets, spectra, ROIs, run records, stretches, library members…), edges = "depends on". Datasets (and in-memory libraries) are the only roots that require a *user decision*; everything else's saveability is a pure function of which roots are saved.
2. **Confirm acyclicity** and expose the roots and each item's transitive dependencies. (Per analysis, the graph is a DAG with datasets as sinks; the resolver should assert this and fail loudly if a cycle ever appears.)
3. **Compute the cascade**: given a set of unsaved roots, classify every other item as one of:
   - **FAITHFUL** — all deps present → reconstruct the real object on load.
   - **SNAPSHOT** — a dep is cut but the item is cheap to freeze (e.g. a dataset-backed spectrum → `NumPyArraySpectrum`) → save the data self-contained.
   - **DROP** — cannot snapshot or user declined → omit, and record a warning.
4. **Provide the policy decision per item type** (which edges are snapshot-able vs. dropped) so the user facing save project file dialog and the object that writes the project file share the same knowledge ion one place.
5. **Produce a human-readable report** of the cascade (what will be faithful / snapshotted / dropped and why) for the save project file dialog (#626) to render.

This engine encodes the epic's governing rule: *faithful when deps present, snapshot when cut, drop only if declined — never a dangling reference.*

Centralizing the dependency logic guarantees consistent, explainable save behavior — the user always knows exactly what will and won't survive — supports WISER's emphasis on reproducible, trustworthy analysis.

- **Let each persister decide independently.** Rejected: duplicated logic, inconsistent edge cases, high risk of dangling references.
- **Always snapshot everything (no graph).** Rejected: wasteful and loses liveness; snapshot is a fallback, not a default.
- **Refuse to save unless every dependency is saved.** Rejected: too rigid — the run histories were explicitly designed to outlive their datasets ("closed runs").

## Additional context

- Depends on the format/pyrep conventions in #616.
- Consumed by the save dialog #626 and informs every item persister (05–08 especially).
- The snapshot pattern mirrors how run records already snapshot endmembers/eigenvalues — small, self-contained values saved directly into the record ([app-state.md](https://github.com/Ehlmann-research-group/WISER/blob/094187c1/doc/sphinx-general-wiser-docs/source/developer-content/app-state.md)).

mmaclay:
## What does this change do?
Adds the dependency resolver / save-policy engine (#617) and the active + collected spectra persister (#620). The resolver is the single place that decides, per saveable item, whether it is reconstructed live, frozen to a snapshot, or dropped; spectra are its first consumer.

- [x] Feature
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] Style
- [ ] Code Refactor
- [x] Test
- [ ] Hot Fix
- [ ] Build
- [ ] CI
- [ ] Chore
- [ ] Revert

## Why is this change needed?
WISER's saveable state is a dependency DAG rooted at datasets. When a dataset isn't saved, everything that transitively depends on it becomes unsaveable — without one component that understands these edges, each persister would re-implement "is this saveable?" ad hoc and risk writing a dangling reference. Spectra are the first state with a real cut edge (a dataset-backed spectrum holds a live dataset), so they're the resolver's first consumer.

Closes #617 (resolver) and #620 (spectra). Part of epic #615.

**Stacked on #618** (`feat/proj-618-datasets`, PR #665). This PR is based on that branch, so its diff shows only the resolver + spectra changes — review/merge #664 then #665 first, after which this rebases onto main.

- **`project/resolver.py`** (#617) — `DependencyResolver.classify(deps, snapshotable)` returns `FAITHFUL` (all deps saved), `SNAPSHOT` (a dep is cut but freezable), or `DROP`. ROIs are standalone `[SOURCE]` (always saved), so for spectra the only cut edge is an unsaved dataset. Persisters declare their own dependencies and consult the shared rule (keeps the resolver stable as new item types are added); `cascade_report` produces the structured summary the save dialog (#626) will render.
- **`project/persisters/spectra.py`** (#620) — persists `_collected_spectra` (ordered) and `_active_spectrum` (a separate entry; the two are mutually exclusive in app_state). Each spectrum serializes as one of three `kind`s:
  - `numpy` — self-contained values + wavelengths(+units) + bad-band mask + name/color/source; used for `NumPyArraySpectrum` **and** for any snapshot.
  - `raster-backed` — `SpectrumAtPoint` by `dataset_id` + `point` + `area` + `avg_mode`.
  - `roi-average` — `ROIAverageSpectrum` by `dataset_id` + `roi_id` + `avg_mode`.
  Dataset-backed spectra reconstruct live when their dataset is saved and freeze to a `numpy` snapshot when it's cut; a faithful reference whose dataset/ROI is missing on load is dropped, not dangled. Spectrum ids aren't preserved (`_all_spectra` is `[DERIVED]`).

Two scoping calls: the resolver's session-wide graph + acyclicity check is **deferred** until run-records/stretches create a real multi-level graph (the current graph is trivially acyclic; the shared classify rule + report ship now); spectrum value/wavelength arrays are **inlined** in the manifest (1-D, small) rather than `.npy` sidecars.

## Added tests?
- [x] yes

`test_project_resolver.py` (classification: faithful / snapshot / drop, ROI defaults + constraint, cut-edge detection, unknown-kind error, report shape) and `test_project_spectra.py` (numpy round-trip, `SpectrumAtPoint` + `ROIAverageSpectrum` faithful round-trips, snapshot-when-dataset-unsaved, active spectrum, drop-on-missing-dataset). **34 project-file tests green** (15 new + the existing 19).

## Added to documentation?
- [x] no documentation needed

## Checklist
- [x] There is an issue associated with this pull request (#617, #620)
- [ ] Branch is up-to-date with main/master — stacked on `feat/proj-618-datasets`; rebases onto main after #664/#665 land

- Merge order is #664 → #665 → this. After #665 lands, re-parent/rebase this onto main.
- Deferred follow-ups: the resolver's graph + acyclicity guard (with #622/#623), and library-member spectra #621 which reuses this snapshot path.

### DIFF
--- a/doc/sphinx-general-wiser-docs/source/developer-content/save-to-project.md
+++ b/doc/sphinx-general-wiser-docs/source/developer-content/save-to-project.md
@@ -118,7 +118,7 @@ types, and bad enum values degrade gracefully.
 
 | Persister | Persists | Notes |
 |-----------|----------|-------|
-| `datasets` | Loaded datasets | File-backed by reference; RAM-backed to ENVI sidecars under `datasets/`. Original ids preserved on load so references stay valid. |
+| `datasets` | Loaded datasets | File-backed by reference, RAM-backed to ENVI sidecars under `datasets/`; a NetCDF subdataset also records which subdataset was selected, and a JSON snapshot of the runtime-editable metadata (data-ignore, bad bands, wavelengths + units, band names, display bands, georeferenced CRS) rides in the manifest and is reapplied on load. Original ids preserved so references stay valid. |
 | `rois` | Regions of interest | Standalone; multi-selection geometry round-trips through the pyrep convention. |
 | `spectra` | Collected + active spectra | Three kinds: self-contained NumPy, raster-backed (dataset + point + area), ROI-average (dataset + roi). Dataset-backed spectra go faithful when the dataset is saved, freeze to NumPy when cut. |
 | `stretches` | Per-`(dataset, band)` contrast stretches | Dataset-cascade leaf: dropped when its dataset is unsaved. Polymorphic across the stretch types. |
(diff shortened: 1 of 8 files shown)


Joooshe (dismissed):
Looks good. If documentation needs to be updated that would be good to do before merging in.

PR #672, opened 2026-07-07
feat(project): spectral-library persister (#621)

### LINKED ISSUE #621: [Project File] Spectral library persistence

mmaclay:
## What does this change do?
Persists spectral libraries (`ApplicationState._spectral_libraries`) as project state, reusing the per-member spectrum rule from #620. A library needs no dependency logic of its own — it's just a named list of spectra, and each member is saved/loaded by the same FAITHFUL/SNAPSHOT/DROP rule as collected/active spectra.

- [x] Feature
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] Style
- [ ] Code Refactor
- [x] Test
- [ ] Hot Fix
- [ ] Build
- [ ] CI
- [ ] Chore

## Why is this change needed?
Spectral libraries are `[SOURCE]` state lost on close. A `ListSpectralLibrary` is just a `List[Spectrum]`, so a single in-memory library can be heterogeneous (self-contained `NumPyArraySpectrum` mixed with dataset-backed members) — a naive whole-library snapshot would mishandle the dataset-backed ones. Reusing #620's per-member serialization resolves that: the decision was always per-spectrum.

Closes #621. Part of epic #615.

**Stacked on #620** (`feat/proj-620-spectra`, PR #670), which it reuses per member. This PR is based on that branch, so its diff shows only the library persister — review/merge #664 → #665 → #670 first, after which this rebases onto main.

New `project/persisters/libraries.py` — `save_libraries` / `load_libraries` + `library_to_pyrep` / `library_from_pyrep`, importing `spectrum_to_pyrep` / `spectrum_from_pyrep` from the #620 persister rather than duplicating spectrum logic. Two storage kinds:
- **`reference`** — an `ENVISpectralLibrary` records only its `.sli`/`.hdr` path and is re-opened from it on load (a large standard library such as the USGS mineral library never enters the bundle); a path that has since moved drops the library and is reported, mirroring file-backed datasets (#618).
- **`inline`** — an in-memory `ListSpectralLibrary` snapshots every member into the manifest; members are self-contained `NumPyArraySpectrum` values, so the restored library is fully self-contained.

**One divergence from the issue text, flagged for the epic doc:** #621 specifies "`_path` set → reference, `_path is None` → inline." The code doesn't support that literally — a `ListSpectralLibrary` is constructed with `path=` set to the *source-import* file (a text/FITS/CSV the spectra came from, e.g. `spectra_table.py`, `app.py`), which is **not re-openable as a library**. Keying on `_path` would try to re-open a text file as an ENVI library and fail. So reference-vs-inline keys on **type** (`ENVISpectralLibrary` → reference; in-memory library → inline) instead — the correct discriminator for the issue's intent (big standard libraries referenced, curated in-memory libraries snapshotted).

Smaller calls: the persister accepts a `resolver` and passes it per-member (self-contained members are FAITHFUL today, but a dataset-backed member would snapshot correctly when its dataset is cut — future-proof); libraries are registered with a **fresh id on load** since nothing references a library by id (unlike datasets), so no id-preservation plumbing is needed.

## Added tests?
- [x] yes

`test_project_libraries.py`: inline round-trip (members restore as self-contained `NumPyArraySpectrum` with values/names/description intact), ENVI reference round-trip (re-opened from its path using the in-package USGS library as a fixture), reference dropped when the file is missing, and unknown storage kind dropped. Full project suite (libraries + spectra + resolver + bundle) green; ruff + ruff-format clean.

## Added to documentation?
- [x] no documentation needed

## Checklist
- [x] There is an issue associated with this pull request (#621)
- [ ] Branch is up-to-date with main/master — stacked on `feat/proj-620-spectra`; rebases onto main after #664/#665/#670 land

- Merge order: #664 → #665 → #670 → this. After #670 lands, re-parent/rebase this onto main.
- Flag to @Joooshe: the epic doc's `_path`-based reference/inline rule for libraries should be updated to the type-based discriminator (a `ListSpectralLibrary`'s `_path` is a source-import file, not a re-openable library).- [ ]

### DIFF
--- a/doc/sphinx-general-wiser-docs/source/developer-content/save-to-project.md
+++ b/doc/sphinx-general-wiser-docs/source/developer-content/save-to-project.md
@@ -118,7 +118,7 @@ types, and bad enum values degrade gracefully.
 
 | Persister | Persists | Notes |
 |-----------|----------|-------|
-| `datasets` | Loaded datasets | File-backed by reference; RAM-backed to ENVI sidecars under `datasets/`. Original ids preserved on load so references stay valid. |
+| `datasets` | Loaded datasets | File-backed by reference, RAM-backed to ENVI sidecars under `datasets/`; a NetCDF subdataset also records which subdataset was selected, and a JSON snapshot of the runtime-editable metadata (data-ignore, bad bands, wavelengths + units, band names, display bands, georeferenced CRS) rides in the manifest and is reapplied on load. Original ids preserved so references stay valid. |
 | `rois` | Regions of interest | Standalone; multi-selection geometry round-trips through the pyrep convention. |
 | `spectra` | Collected + active spectra | Three kinds: self-contained NumPy, raster-backed (dataset + point + area), ROI-average (dataset + roi). Dataset-backed spectra go faithful when the dataset is saved, freeze to NumPy when cut. |
 | `stretches` | Per-`(dataset, band)` contrast stretches | Dataset-cascade leaf: dropped when its dataset is unsaved. Polymorphic across the stretch types. |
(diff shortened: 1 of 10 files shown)


Joooshe (dismissed):
Looks good.

PR #673, opened 2026-07-07
feat(project): contrast-stretch persister (#622)

### LINKED ISSUE #622: [Project File] Contrast-stretch persistence

mmaclay:
## What does this change do?
Persists the committed per-band contrast stretches (`ApplicationState._stretches`, keyed by `(dataset_id, band_index)`) as project state, so a reopened session restores the analyst's chosen visualization instead of falling back to a default stretch on every band. Adds pyrep serialization for the closed stretch hierarchy and a `get_all_stretches()` accessor on `ApplicationState`.

- [x] Feature
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] Style
- [ ] Code Refactor
- [x] Test
- [ ] Hot Fix
- [ ] Build
- [ ] CI
- [ ] Chore
- [ ] Revert

## Why is this change needed?
Stretches are `[SOURCE]` state lost on close; without them a restored session looks visibly wrong. Each stretch is a leaf on the dataset cascade — a stretch can't outlive its dataset (`remove_dataset` deletes a dataset's stretches with it) — so a stretch whose dataset isn't saved is dropped, and on load one whose dataset wasn't restored is dropped rather than left orphaned.

Closes #622. Part of epic #615.

**Stacked on #621** (`feat/proj-621-libraries`). This PR is based on that branch, so its diff shows only the stretch persister — review/merge #664 → #665 → #670 → #621 first, after which this rebases onto main.

New `project/persisters/stretches.py` — `save_stretches` / `load_stretches` + `stretch_to_pyrep` / `stretch_from_pyrep`, storing a flat list `[{dataset_id, band_index, stretch: <pyrep>}]`. On save, a stretch whose dataset isn't saved (resolver) is dropped; on load, a stretch whose dataset wasn't restored is dropped — preserving the "a stretch never outlives its dataset" invariant.

Serialization for the closed hierarchy lives **in the persister** (like the spectra persister), not on the classes — the `…UsingNumba` variants are numba jitclasses that can't host plain methods, and only their parameters are needed. `stretch_to_pyrep` reads the plain fields off whichever variant is held (jitclass fields are Python-readable); `stretch_from_pyrep` rebuilds the **pure-Python** variant, which is behaviorally identical (the JIT variant is a performance choice re-derived when the builder next recomputes the stretch). Seven type tags: `linear` (`_lower`/`_upper`), `equalize`, `decorrelation`, `sqrt`, `log2`, `composite`, `none` (bare `StretchBase`).
- `StretchComposite` (a conditioner wrapping a base stretch) serializes its two halves recursively via `first()`/`second()`.
- `StretchHistEqualize` reconstructs through its **normal constructor**: it retains only the normalized `_cdf` + `_histo_edges`, but because the constructor renormalizes the CDF by its last element, feeding back `diff(cdf) * diff(edges)` as the histogram bins reproduces the identical CDF (verified by a round-trip test on both `_cdf` and `apply()`).
- **`StretchDecorrelation` carries no per-instance state** — it recomputes its joint transform from the data at apply-time — so it persists as just its type tag (this differs from the issue text, which assumed derived state to snapshot; flagged for the epic doc).

Array-based state is **inlined** in the manifest (histogram-sized, 1-D) rather than written to `.npy` sidecars. Out of scope, per the issue: the `StretchBuilder` dialog's working state is `[EPHEMERAL]`/`[DERIVED]` — it rebuilds from the committed stretch on open — so saving the committed stretch is sufficient.

## Added tests?
- [x] yes

`test_project_stretches.py` (7): linear, composite, and histogram-equalize round-trips; the stateless stretches (decorrelation / both conditioners / bare no-stretch); dropped-when-dataset-unsaved (save side); dropped-when-dataset-missing (load side); and malformed-entry-dropped-not-fatal. Full project suite (stretches + libraries + spectra + resolver + bundle) green; ruff + ruff-format clean.

One coverage note: the tests exercise the pure-Python stretch classes (what `load` produces). The save-side reading of a numba jitclass instance's fields is what the app uses in practice but isn't unit-tested here — constructing a JIT variant needs a real dataset over the memory threshold.

## Added to documentation?
- [x] no documentation needed

## Checklist
- [x] There is an issue associated with this pull request (#622)
- [ ] Branch is up-to-date with main/master — stacked on `feat/proj-621-libraries`; rebases onto main after the stack below it lands

- Merge order: #664 → #665 → #670 → #621 → this. After #621 lands, re-parent/rebase this onto main.
- Flag to @Joooshe for the epic doc: `StretchDecorrelation` is stateless (no array snapshot needed), and stretch serialization lives in the persister rather than on the classes (numba jitclasses can't host methods).
                                                                                                                                                           Baked for 30s

### DIFF
--- a/doc/sphinx-general-wiser-docs/source/developer-content/save-to-project.md
+++ b/doc/sphinx-general-wiser-docs/source/developer-content/save-to-project.md
@@ -118,7 +118,7 @@ types, and bad enum values degrade gracefully.
 
 | Persister | Persists | Notes |
 |-----------|----------|-------|
-| `datasets` | Loaded datasets | File-backed by reference; RAM-backed to ENVI sidecars under `datasets/`. Original ids preserved on load so references stay valid. |
+| `datasets` | Loaded datasets | File-backed by reference, RAM-backed to ENVI sidecars under `datasets/`; a NetCDF subdataset also records which subdataset was selected, and a JSON snapshot of the runtime-editable metadata (data-ignore, bad bands, wavelengths + units, band names, display bands, georeferenced CRS) rides in the manifest and is reapplied on load. Original ids preserved so references stay valid. |
 | `rois` | Regions of interest | Standalone; multi-selection geometry round-trips through the pyrep convention. |
 | `spectra` | Collected + active spectra | Three kinds: self-contained NumPy, raster-backed (dataset + point + area), ROI-average (dataset + roi). Dataset-backed spectra go faithful when the dataset is saved, freeze to NumPy when cut. |
 | `stretches` | Per-`(dataset, band)` contrast stretches | Dataset-cascade leaf: dropped when its dataset is unsaved. Polymorphic across the stretch types. |
@@ -126,6 +126,33 @@ types, and bad enum values degrade gracefully.
 | `runs` | PCA / MNF / unmixing / K-Means histories | Every record self-contained; datasets referenced softly by id; run ids re-minted on load. |
 | `crs` | User-created coordinate systems | WKT + creator-dialog state; rebuilt via GDAL/`osr`, bad WKT drops the entry, a bad enum degrades one field. |
 | `bandmath` | Saved band-math expressions | A plain list of expression strings on `ApplicationState`. |
+
+### Dataset metadata and subdatasets
+
+A file-backed dataset is re-opened from its path on load, which re-derives its
+metadata from the source file — so metadata the user edited at runtime (a renamed
+band, an adjusted data-ignore value, a hand-marked bad-band list) would be lost on
+reopen. Each dataset entry therefore also carries a JSON snapshot of the
+runtime-editable metadata: the data-ignore value, the bad-band list, per-band
+wavelengths and their unit, band descriptions, the default display bands, and any
+georeferencer-assigned CRS (WKT) with its geo-transform. On load the snapshot is
+reapplied through the safe per-field setters in a fixed order — data-ignore first,
+since it keys the band-statistics cache, and band descriptions after wavelengths,
+since rebuilding band info overwrites them. Reapply is best-effort and
+field-guarded: a value that no longer fits the reopened dataset (a band-count
+mismatch, an unparseable unit) is skipped so the dataset still restores.
+
+A NetCDF **subdataset** needs one extra field. Its `get_filepaths()` returns a GDAL
+descriptor (`NETCDF:"/path/file.nc":var`) rather than a plain path, so the entry
+records the base `.nc` file together with the `subdataset_name` descriptor and
+reopens via `load_from_file(base, subdataset_name=…, interactive=False)` — restoring
+the *same* subdataset the user had open instead of re-running the auto-pick heuristic
+or dropping the dataset because the descriptor is not a file on disk. Subdataset
+selection is NetCDF-only.
+
+Both the metadata snapshot and `subdataset_name` are **additive, optional** manifest
+fields, so they require no `format_version` bump: an older manifest simply lacks them
+and loads exactly as before.
 
 ---
 
--- a/src/tests/test_project_datasets.py
+++ b/src/tests/test_project_datasets.py
@@ -10,16 +10,21 @@ a malformed id.
 import os
 
 import numpy as np
+from astropy import units as u
 
 import tests.context  # noqa: F401
+
+from osgeo import osr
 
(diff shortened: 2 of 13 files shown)


Joooshe (dismissed):
Looks good. Pretty standard format for these persisters. Good job coming up with it.

PR #679, opened 2026-07-08
feat(project): band-math saved-expression persister (#625)

mmaclay:
## What does this change do?
  Persists the band-math saved-expression list as project state, and moves that list onto `ApplicationState` so it survives the dialog closing (the "nice cleanup" option from the issue).

  ## What type of PR is this? (check all applicable)
  - [x] Feature
  - [ ] Bug Fix
  - [ ] Documentation Update
  - [ ] Style
  - [x] Code Refactor
  - [ ] Performance Improvements
  - [x] Test
  - [ ] Hot Fix
  - [ ] Build
  - [ ] CI
  - [ ] Chore
  - [ ] Revert

  ## Why is this change needed?
  The band-math saved-expression list is `[SOURCE]` state, but it lived in an unusual place — directly in the `BandMathDialog`'s saved-expressions combo-box, not routed through `ApplicationState` — so it was lost on close (unless the user separately exported it to a `.txt`). A session-reconstruction routine would
The band-math saved-expression list is `[SOURCE]` state, but it lived in an unusual place — directly in the `BandMathDialog`'s saved-expressions combo-box, not routed through `ApplicationState` — so it was lost on close (unless the user separately exported it to a `.txt`). A session-reconstruction routine would otherwise have to reach into the dialog to capture it.

Closes #625. Part of epic #615. Depends only on the foundation #616.

**Stacked on #624** (`feat/proj-624-crs`, PR #678). This PR is based on that branch, so its diff shows only the band-math changes — review/merge the stack below it first, after which this rebases onto main.

## How did you implement the change?
- **`ApplicationState`** now owns the list: a `_bandmath_saved_exprs: List[str]` field plus `get_bandmath_expressions()` / `set_bandmath_expressions()`.
- **`BandMathDialog`** treats its combo-box as a view over that store: it **seeds** `cbox_saved_exprs` from `ApplicationState` on open, and **syncs back** via a one-line `_sync_saved_exprs_to_app_state()` helper at the combo's two mutation points (`_on_add_expr_to_saved`, `_on_load_saved_exprs`). The existing `.txt` import/export and the `_saved_exprs_modified` unsaved-changes prompt are left untouched.
- **`project/persisters/bandmath.py`** — `save_bandmath` writes `manifest["bandmath_expressions"]` as a plain string list; `load_bandmath` restores it, leaving the store untouched when the section is absent and dropping/reporting any non-string entries in a malformed list.

[...]

Issue #628, opened 2026-06-23
[Project File] Project versioning & migration


Joooshe:
## Is your feature request related to a problem? Please describe.

A project file format that is saved by users *will* change as WISER evolves — new state items, renamed fields, restructured records. Without an explicit, enforced versioning strategy, one of two failures is inevitable:

- **Silent rot:** a schema change quietly breaks the ability to open older project files, and nobody notices until a user reports a corrupted/unopenable project.
- **Accreting cruft:** each change adds ad-hoc branching to the load path until it is unmaintainable.

The bundle format (#616) provides a `format_version` field and a migration *seam*, and the load orchestrator (#627) checks the version — but neither commits to a **policy** or, crucially, **tests that old files still load**. This issue owns that policy and its enforcement.

## Describe the solution you'd like

A **versioning & migration subsystem** with a guarantee that is actually tested.

**Guarantee:** *Any project file written by a released WISER can be opened by every later WISER.* (Backward compatibility.) Forward compatibility is explicitly **not** promised.

**Components:**

1. **Migrate-up chain.** One loader understands only the **current** schema. On load, the `format_version` is read and a chain of small *pure* functions (`migrate_v1_to_v2`, `migrate_v2_to_v3`, …) transforms the manifest pyrep dict step-by-step up to the current version before the normal load runs. Sub-issues 03–12 never see old shapes. Each migration is independently unit-tested with a minimal before/after dict.
2. **Refuse-too-new behavior.** If a file's `format_version` is **greater** than the running WISER understands, abort the load with a clear, user-facing message ("This project was created with a newer version of WISER. Please upgrade to open it.") — never a crash, partial load, or silent data loss.
3. **Additive-change convention (documented).** Non-breaking additions (new *optional* fields) require **no** version bump and **no** migration: `from_pyrep` ignores unknown keys and defaults missing ones. The `format_version` is bumped (and a migration written) **only** for breaking changes — renamed, removed, restructured, or semantically-changed fields. Document this rule where contributors will see it.
4. **Golden-file regression suite (the enforcement).** Check in a real, representative project bundle for **every historical `format_version`** under test fixtures. A test loads each on the current code and asserts a successful, correct restore. This is what converts the guarantee from aspiration into something CI fails on. Add a checklist/CI reminder so that whenever `format_version` is bumped, a new golden fixture for the previous version is captured.
5. **A "too-new" fixture** verifying the refuse-cleanly path.

## Describe alternatives you've considered

[...]

Issue #625, opened 2026-06-23
[Project File] Band-math saved-expression persistence


Joooshe:
## Is your feature request related to a problem? Please describe.

The band-math **saved-expression list** is `[SOURCE]` state, but it lives in an unusual place: directly in the combo-box widget `_ui.cbox_saved_exprs` on the `BandMathDialog` ([bandmath_dialog.py](https://github.com/Ehlmann-research-group/WISER/blob/094187c1/src/wiser/gui/bandmath_dialog.py)), with `_saved_exprs_modified` tracking unsaved changes. It is **not** routed through `ApplicationState`, so a session-reconstruction routine has to reach into the dialog to capture it. Today these saved expressions are lost on close (unless the user separately exported them to a `.txt`).

## Describe the solution you'd like

Persist the saved-expression list as part of the project.

**Save:** capture the list of expression strings from the dialog's saved-expression store (the combo-box items) into the manifest — a simple list of strings. Independent of datasets and every other item.

**Load:** repopulate the saved-expression store on the `BandMathDialog` (creating/seeding it so the expressions are present whether or not the dialog has been opened yet).

**Out of scope** (`[EPHEMERAL]` / not session state): the current expression text in `_ui.ledit_expression`, variable bindings (`tbl_variables`), transient `_expr_info`, and batch-job definitions. **Old band-math runs are explicitly not tracked**; band-math *output rasters* are datasets, handled by #618.

## Describe alternatives you've considered

- **Rely on using just the existing `.txt` export/import.** Rejected: manual, separate from the session, easily forgotten.
- **Move the saved-expression list onto `ApplicationState` first.** A nice cleanup but not required; the persister can read/write the dialog's store directly. Could be a follow-up.

## Additional context

- Depends only on #616.
- Note the architectural wrinkle (state on a dialog, not `ApplicationState`), shared with the georeferencer table — see [app-state.md](https://github.com/Ehlmann-research-group/WISER/blob/094187c1/doc/sphinx-general-wiser-docs/source/developer-content/app-state.md).

PR #680, opened 2026-07-08
feat(project): save/load orchestrator (#627, #626 save flow)

### LINKED ISSUE #626: [Project File] Save flow + dependency-aware dialog

### LINKED ISSUE #627: [Project File] Load / restore orchestration

Saving is only half the feature. Opening a project must **rebuild the session in the correct order** so that every reference resolves: datasets must exist (with their original ids) before stretches, spectra, and run records that reference them are restored; ROIs must exist before ROI-average spectra are made live again. Done wrong, the load produces cross-wired references, missing-dependency crashes, or silently inert state.

A **load/restore orchestrator** that opens a bundle and reconstructs the session in dependency (topological) order, on a **cleared session** (no merge — per epic scope).

**Order of operations:**
1. Open the bundle (#616); read `manifest.json`; run the **migrate-up chain** to bring an older `format_version` to the current schema, or **refuse cleanly** if the file is newer than this WISER understands (both owned by #628). After this step the rest of the load only ever sees the current schema.
2. **Clear the current session** to a clean state.
3. Restore **datasets** first, **preserving original ids**, then set `_next_id = max(id)+1` (#618).
4. Restore **user CRSs** #624 and **band-math expressions** #625 (independent — any time).
5. Restore **ROIs** #619.
6. Restore **stretches** #622, keyed onto restored datasets; drop leaves whose dataset is absent.
7. Restore **spectra** #620 and **spectral libraries** #621: rebuild FAITHFUL objects against restored datasets/ROIs; otherwise rehydrate frozen snapshots as `NumPyArraySpectrum` instances.
8. Restore **run histories** #623: wire records to restored datasets, or place into "closed runs" when a dataset is absent.
9. Rebuild **`_all_spectra`** — the by-id index `{spectrum_id: Spectrum}` that other code uses to look up spectra by id. It is not saved (it duplicates the active/collected collections), but it must be repopulated after step 7 so that any id-based lookup resolves to the freshly-restored objects. The three other `[DERIVED]` items require no action: `DataCache` (render/computation/histogram caches), `RasterDataSet._cached_band_stats` (per-band min/max), and each spectrum's `_icon` thumbnail all refill themselves the first time something asks for them.
10. **Emit the change signals so the UI reflects the restored session.** WISER's UI is signal-driven: `ApplicationState` emits a Qt signal on each change and the panels (layer list, spectrum plot, main view) redraw in response. Because restore loads everything programmatically into a cleared (empty-UI) session, each restored item needs its corresponding signal fired or the data sits in `ApplicationState` with nothing on screen. Wrap this in a single helper (e.g. `emit_session_reload_signals()`) that fires the **granular** per-item signals ([app_state.py:73-104](https://github.com/Ehlmann-research-group/WISER/blob/094187c1/src/wiser/gui/app_state.py#L73-L104)):
   - Datasets → `dataset_added(id, bool)` per dataset, then `mainview_dataset_changed(id)` for the one being displayed.
   - Spectral libraries → `spectral_library_added(id)`.
   - ROIs → `roi_added(RegionOfInterest)`.
   - Stretches → `stretch_changed(dataset_id, bands_tuple)`.
   - Active spectrum → `active_spectrum_changed()`; collected spectra → `collected_spectra_changed(object, int, int)`.

   Two things this step must also handle, because they do **not** go through `ApplicationState` signals:
   - **Run histories** signal through their own managers, not `ApplicationState` — fire each `RunHistoryManagerBase.records_changed` (PCA / MNF / linear-unmixing / K-Means). (The legacy `kmeans_centroids_changed` goes away with [issue #613](https://github.com/Ehlmann-research-group/WISER/issues/613).)
   - **User CRSs and band-math expressions** have no `ApplicationState` signal at all — they live on their dialogs and are refreshed by repopulating those dialogs' stores directly (steps 4).

   **Granular vs. bulk note:** we deliberately choose the granular approach — reusing the same per-item signals that `add_dataset()` and friends already emit, just gathered behind `emit_session_reload_signals()`. There is no single "whole session reloaded" signal today, and we are not adding one right now. If a bulk restore ever produces a signal flurry that is too slow or causes visible flicker, introducing one coarse "session reloaded, refresh everything" signal would be a small, contained follow-up — but granular is the default and is sufficient for now.

**Robustness:** missing sidecar files, an unreadable file-backed dataset path, or an unknown `type` tag should degrade gracefully (warn + continue), never hard-crash the load. Produce a load report mirroring the save report (what restored faithfully / as snapshot / was dropped) and log.

- **Restore in manifest/declaration order.** Rejected: references may not yet exist; topological order is required.
- **Merge into the existing session.** Out of scope (epic): open always clears first, which is what makes id-preservation safe.
- **Hard-fail on any missing dependency.** Rejected: graceful degradation (snapshots, closed runs, warnings) is the whole point of the design.

## Additional context

- Depends on #616, #618, all item persisters #619–#625, and the versioning/migration subsystem #628; pairs with the save side #626.
- The preserve-ids-on-fresh-load approach (no remapping) is the key invariant that keeps every restored reference valid.

mmaclay:
## What does this change do?
  Adds the top-level save/load orchestration that ties the per-item persisters into a working feature — the first code that writes and reopens a real `.wiserproj` end-to-end. A full session round-trips: dataset + ROI + spectrum + stretch + user CRS + band-math expressions + run records, save to disk and reopen into a cleared session.

  - [x] Feature
  - [ ] Bug Fix
  - [ ] Documentation Update
  - [ ] Style
  - [ ] Code Refactor
  - [x] Test
  - [ ] Hot Fix
  - [ ] Build
  - [ ] CI
  - [ ] Chore
  - [ ] Revert

  ## Why is this change needed?
  Every per-item persister (#618–#625) was only unit-tested in isolation — nothing wrote a real bundle or reopened one. Saving is only half the feature; opening a project must rebuild the session in dependency order so every reference resolves (datasets must exist, with their original ids, before the stretches/spectra/run records that reference them; ROIs before ROI-average spectra). Done wrong, a load produces cross-wired references or missing-dependency crashes.

  Closes #627 (load/restore orchestration). Implements the **save-flow half of #626** — the dependency-aware Qt Save dialog and menu wiring remain as a follow-up. Depends on the foundation #616, the resolver #617, and all item persisters #618–#625.

  **Stacked on #625** (`feat/proj-625-bandmath`, PR #679). This PR is based on that branch, so its diff shows only the orchestrator — review/merge the stack below it first, after which this rebases onto main.

  New `project/orchestrate.py`:
  - **`save_project(app_state, dest, resolver=None)`** — writes every item into one manifest (plus dataset/array sidecars via the bundle) by calling the eight `save_*` in dependency order (datasets first — they're the roots and own the sidecar I/O). A `dest` ending in `.wiserproj` is packaged as a zip (built in a temp dir); any other path is a bundle directory. Without an explicit resolver, every dataset is treated as saved (the #626 dialog will supply a user-driven one).
  - **`load_project(src, app_state, extract_dir=None)`** — opens a bundle (unzipping a `.wiserproj` into `extract_dir`, which must outlive the session since  one).
- **`load_project(src, app_state, extract_dir=None)`** — opens a bundle (unzipping a `.wiserproj` into `extract_dir`, which must outlive the session since sidecar datasets are read lazily), calls `bundle.read_manifest()` (which already migrates up / raises `ProjectTooNewError`), **clears the session**, then restores in topological order: datasets → CRS/band-math → ROIs → stretches → spectra → libraries → runs. Returns a load report of the entries that couldn't be restored, per section.

Two audit findings made this simpler than the issue anticipated: the UI reload signals **fire inline** (each `load_*` restores through the signal-emitting `add_*`/`set_*` accessors), and `_all_spectra` **self-rebuilds** via `collect_spectrum`/`set_active_spectrum` — so no separate `emit_session_reload_signals`/rebuild-index pass is needed.

Also adds two small shared-state pieces:
- **`ApplicationState.clear_session()`** — resets to an empty session using the normal `remove_*` methods so removed-signals fire and the UI clears. Verified safe: `remove_dataset → delete_underlying_dataset → delete_dataset` only `driver.Delete`s scratch `IN_DISK_NOT_SAVED` temp rasters — it never touches a user's source file.
- **`RunHistoryManagerBase.clear_records()`**.

Main-view dataset selection (which dataset is displayed) is derived UI state on the MainView widget, not `[SOURCE]` app_state, so it isn't persisted here — the main view defaults to `add_dataset`'s view pick on load. Restoring the exact saved view is a small follow-up if wanted.

## Added tests?
- [x] yes

`test_project_orchestrate.py` (3): a full-session round-trip through a real `.wiserproj` **zip** and through a bundle **directory** (asserting the dataset restores with its original id and its pixels, plus the ROI / spectrum / stretch / CRS / band-math / PCA record), and a **clear-session-first** test (a load discards unrelated state already in the destination). Full project suite green — 66 passed, no regression from the `app_state.py`/`run_history.py` additions; ruff + ruff-format clean.

One coverage note: the round-trip is proven via a comprehensive fake `ApplicationState` (matching the pattern the other persister tests use — a real Qt `ApplicationState` needs an app + cache). The real `clear_session` is mechanical and exercised through the app rather than unit-tested here.

## Added to documentation?
- [x] no documentation needed

## Checklist
- [x] There is an issue associated with this pull request (#627, #626)
- [ ] Branch is up-to-date with main/master — stacked on `feat/proj-625-bandmath`; rebases onto main after the stack below it lands

- Merge order: the persister stack merges bottom-up; after #679 lands, re-parent/rebase this onto main.
- Remaining epic work: the #626 dependency-aware Qt Save dialog + Save/Save-As menu wiring, and the #628 golden-file version gate.

### DIFF
--- a/src/wiser/gui/app_state.py
+++ b/src/wiser/gui/app_state.py
@@ -888,8 +888,23 @@ class ApplicationState(QObject):
         with no removal path (CRSs, band-math expressions, run records) are
         emptied directly.
         """
-        for ds_id in list(self._datasets.keys()):
-            self.remove_dataset(ds_id)
+        managers = (
+            self._pca_history,
+            self._mnf_history,
+            self._linear_unmix_history,
+            self._kmeans_history,
+        )
+        # The run-history managers react to dataset_removed; block their signals
+        # during the bulk removal so each emits records_changed once (from
+        # clear_records below) rather than once per removed dataset.
+        for manager in managers:
+            manager.blockSignals(True)
+        try:
+            for ds_id in list(self._datasets.keys()):
+                self.remove_dataset(ds_id)
+        finally:
+            for manager in managers:
+                manager.blockSignals(False)
         for roi_id in list(self._regions_of_interest.keys()):
             self.remove_roi(roi_id)
         for lib_id in list(self._spectral_libraries.keys()):
@@ -900,12 +915,7 @@ class ApplicationState(QObject):
         self._all_spectra.clear()
         self._user_created_crs.clear()
         self._bandmath_saved_exprs = []
-        for manager in (
-            self._pca_history,
-            self._mnf_history,
-            self._linear_unmix_history,
-            self._kmeans_history,
-        ):
+        for manager in managers:
             manager.clear_records()
         self._next_id = 1
 
--- a/src/wiser/gui/app_state.py
+++ b/src/wiser/gui/app_state.py
@@ -672,10 +672,15 @@ class ApplicationState(QObject):
         """
         self._config.set(option, value)
 
-    def add_roi(self, roi: RegionOfInterest, make_name_unique=False) -> None:
+    def add_roi(self, roi: RegionOfInterest, make_name_unique=False, roi_id: Optional[int] = None) -> None:
         """
         Add a Region of Interest to WISER's state.  A ``ValueError`` is raised
         if the ROI does not have a unique name.
+
+        When ``roi_id`` is given (project restore), that id is used verbatim
+        instead of allocating a new one, and the internal id counter is advanced
+        past it so later allocations do not collide.  Preserving the id keeps
+        ROI-average spectra, which reference their ROI by id, resolvable on load.
         """
 
         if roi is None:
@@ -713,7 +718,12 @@ class ApplicationState(QObject):
             color = get_random_matplotlib_color(colors_in_use)
             roi.set_color(color)
 
-        roi_id = self.take_next_id()
+        if roi_id is None:
+            roi_id = self.take_next_id()
+        elif roi_id in self._regions_of_interest:
+            raise ValueError(f"ROI id {roi_id} is already in use")
+        else:
+            self._next_id = max(self._next_id, roi_id + 1)
         roi.set_id(roi_id)
         self._regions_of_interest[roi_id] = roi
         self.roi_added.emit(roi)
--- a/src/wiser/gui/app_state.py
+++ b/src/wiser/gui/app_state.py
@@ -878,6 +878,37 @@ class ApplicationState(QObject):
     def set_bandmath_expressions(self, expressions: List[str]) -> None:
         self._bandmath_saved_exprs = list(expressions)
 
+    def clear_session(self) -> None:
+        """Reset to an empty session, discarding all current state.
+
+        Used before restoring a project file (the load orchestrator clears first
+        so that datasets can be restored with their original ids).  Removal goes
+        through the normal ``remove_*`` methods so the corresponding
+        removed-signals fire and the UI clears; the remaining ``[SOURCE]`` stores
+        with no removal path (CRSs, band-math expressions, run records) are
+        emptied directly.
+        """
+        for ds_id in list(self._datasets.keys()):
+            self.remove_dataset(ds_id)
+        for roi_id in list(self._regions_of_interest.keys()):
+            self.remove_roi(roi_id)
+        for lib_id in list(self._spectral_libraries.keys()):
+            self.remove_spectral_library(lib_id)
+        self.remove_all_collected_spectra()
+        self.set_active_spectrum(None)
+        self._stretches.clear()
+        self._all_spectra.clear()
+        self._user_created_crs.clear()
+        self._bandmath_saved_exprs = []
+        for manager in (
+            self._pca_history,
+            self._mnf_history,
+            self._linear_unmix_history,
+            self._kmeans_history,
+        ):
+            manager.clear_records()
+        self._next_id = 1
+
     def get_user_created_crs(self):
         return self._user_created_crs
 
--- a/src/tests/test_project_bundle.py
+++ b/src/tests/test_project_bundle.py
@@ -103,6 +103,19 @@ def test_bundle_array_sidecar(tmp_path):
     np.testing.assert_array_equal(bundle.fetch_array(ref), arr)
 
 
+def test_bundle_clear_contents_removes_stale_sidecars(tmp_path):
+    bundle = ProjectBundle.create(tmp_path / "proj")
+    bundle.add_array("centroids", np.arange(4, dtype=np.float32))
+    bundle.write_manifest({"rois": []})
+    assert (bundle.root / ProjectBundle.ARRAYS_DIR).is_dir()
+    assert (bundle.root / ProjectBundle.MANIFEST_NAME).is_file()
+
+    bundle.clear_contents()
+    assert not (bundle.root / ProjectBundle.ARRAYS_DIR).exists()
+    assert not (bundle.root / ProjectBundle.MANIFEST_NAME).exists()
+    assert bundle.root.is_dir()  # the bundle directory itself remains
+
+
 def test_bundle_zip_round_trip(tmp_path):
     bundle = ProjectBundle.create(tmp_path / "proj")
     manifest = {"rois": [roi_to_pyrep(_sample_roi())]}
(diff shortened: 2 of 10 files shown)


Joooshe:
I think this is duplicated in your original comment "load_project(src, app_state, extract_dir=None) — opens a bundle (unzipping a .wiserproj into extract_dir, which must outlive the session since one)."

src/wiser/gui/app_state.py
```diff
@@ -878,6 +878,47 @@ def get_bandmath_expressions(self) -> List[str]:
...
+        for roi_id in list(self._regions_of_interest.keys()):
+            self.remove_roi(roi_id)
+        for lib_id in list(self._spectral_libraries.keys()):
+            self.remove_spectral_library(lib_id)
+        self.remove_all_collected_spectra()
+        self.set_active_spectrum(None)
+        self._stretches.clear()
+        self._all_spectra.clear()
+        self._user_created_crs.clear()
+        self._bandmath_saved_exprs = []
```
Joooshe, line 927: I may be wrong but it doesn't seem like the band math dialog's saved expressions would be cleared by doing this.

Joooshe: Ahhhh, I see, at start up the reference system creator (which self._user_created_crs.clear() affects) and the band math dialog both sychronize with app state at start up so the UI is in sync with the backend. That looks good then.

Joooshe: I think the main thing I was worried about is that there is nothing right now that tests to make sure that app_state changes stay sychronized with the CRSs in the reference system creator or the band math expressions in band math. I made an issue about this in #689 but that is outside the scope of this PR.

Joooshe (dismissed):
There is a lot going on here. It looks good though. It will be very important that when the feature is done we make a e2e GUI test for it and test it manually.

PR #681, opened 2026-07-08
feat(project): dependency-aware Save Project dialog + menu wiring (#626)

### LINKED ISSUE #626: [Project File] Save flow + dependency-aware dialog

mmaclay:
## What does this change do?
Completes the Save flow with a dependency-aware Save Project dialog and the File-menu wiring (Open Project / Save Project / Save Project As). The dialog lets the user choose which in-memory datasets to include and previews the consequences live before writing.

- [x] Feature
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] Style
- [ ] Code Refactor
- [x] Test
- [ ] Hot Fix
- [ ] Build
- [ ] CI
- [ ] Chore
- [ ] Revert

## Why is this change needed?
Even with the persisters and orchestrator built, the user needs to understand and control what gets saved — especially which in-memory datasets to include, since those are the only roots that require a decision and everything else cascades from them. Without a clear UI, users would unknowingly drop dataset-backed spectra or stretches when they skip a dataset, and would not understand why something "didn't save."

Closes #626. Part of epic #615. Builds on the resolver #617, the orchestrator's save flow (#627), and all item persisters.

**Stacked on the orchestrator** (`feat/proj-627-orchestrate`, PR #680). Its diff shows only the dialog + wiring — review/merge the stack below it first, after which this rebases onto main.

- **`project/save_plan.py`** — the dialog's headless, unit-tested model: `savable_dataset_roots` (RAM- vs file-backed split — file-backed datasets are referenced for free, so only RAM-backed ones are decision points), `resolver_for_selection` (build a `DependencyResolver` from the user's exclusions), and `save_plan` (the live FAITHFUL/SNAPSHOT/DROP cascade — dataset-backed spectra that will freeze to a snapshot, stretches that will drop; runs and library members are self-contained and don't cascade).
- **`gui/save_project_dialog.py`** — a hand-built modal `SaveProjectDialog` (a checkable `QTableWidget` of RAM-dataset roots + a consequences table refreshed live on toggle, via `save_plan`), returning the chosen resolver from `get_resolver()`. Hand-built rather than a generated `Ui_*` to avoid the `.ui`/`pyside6-uic` toolchain.
- **`gui/app.py`** — File-menu actions **Open Project**, **Save Project** (`Ctrl+S`), **Save Project As** (`Ctrl+Shift+S`) with handlers that drive `save_project` / `load_project`; a `_current_project_path` field so Save re-saves in place; Open Project confirms before discarding the session and warns about any dropped items on load; a too-new file is reported cleanly (`ProjectTooNewError`).

Scoped-out for v1 (follow-ups, not blocking): excluding in-memory *libraries* from the dialog (all libraries currently save); Save As is two dialogs (options then file picker); a zipped project opens by extracting to a temp dir the app should later own/clean; `Ctrl+S` re-saves everything rather than reusing the last exclusion set.

## Added tests?
- [x] yes

`test_project_save_plan.py` (4): the RAM/file split, all-faithful when nothing is excluded, snapshot+drop when a dataset is excluded, and a self-contained spectrum staying out of the plan. `test_project_save_dialog.py` (3): the dialog's resolver reflects the checkbox state, a file-backed dataset is not a checkable root (constructed under a headless `QApplication`), and an `import wiser.gui.app` smoke test for the menu wiring. Full project suite green (78 passed); ruff + ruff-format clean.

## Added to documentation?
- [x] no documentation needed

## Checklist
- [x] There is an issue associated with this pull request (#626)
- [ ] Branch is up-to-date with main/master — stacked on `feat/proj-627-orchestrate`; rebases onto main after the stack below it lands

- Merge order: the stack merges bottom-up; after #680 lands, re-parent/rebase this onto main.
- Follow-ups: library exclusion in the dialog, a single-step Save As, a session-owned extract dir for zipped projects, and `Ctrl+S` reusing the last resolver.
- The remaining epic work is #628 (versioning), stacked on this.

### DIFF
--- a/src/tests/test_project_save_dialog.py
+++ b/src/tests/test_project_save_dialog.py
@@ -1,0 +1,82 @@
+"""Tests for the Save-Project dialog wiring (issue #626).
+
+The dialog's dependency logic lives in and is tested through
+``wiser.project.save_plan``; here we only check the thin Qt glue -- that a
+RAM-backed dataset becomes a checkable root, a file-backed one does not, and the
+resolver returned by the dialog reflects the checkbox state -- plus a smoke
+import of ``wiser.gui.app`` to catch the menu-wiring edits.
+"""
+
+import tests.context  # noqa: F401
+
+from PySide6.QtCore import Qt
+from PySide6.QtWidgets import QApplication
+
+from wiser.gui.save_project_dialog import SaveProjectDialog
+from wiser.project.resolver import Dependency
+
+
+def _qapp():
+    return QApplication.instance() or QApplication([])
+
+
+class _FakeDataset:
+    def __init__(self, ds_id, filepaths, name):
+        self._id = ds_id
+        self._filepaths = filepaths
+        self._name = name
+
+    def get_id(self):
+        return self._id
+
+    def get_filepaths(self):
+        return self._filepaths
+
+    def get_name(self):
+        return self._name
+
+
+class _FakeAppState:
+    def __init__(self, datasets):
+        self._datasets = datasets
+
+    def get_datasets(self):
+        return list(self._datasets)
+
+    def get_collected_spectra(self):
+        return []
+
+    def get_active_spectrum(self):
+        return None
+
+    def get_all_stretches(self):
+        return {}
+
+
+def test_dialog_resolver_reflects_selection():
+    _qapp()
+    app_state = _FakeAppState([_FakeDataset(1, [], "cube")])
+    dialog = SaveProjectDialog(app_state, None)
+
+    # Checked by default -> the dataset is saved.
+    assert dialog.get_resolver().is_saved(Dependency("dataset", 1))
+
+    # Unchecking the root excludes it from the resolver.
+    dialog._roots_table.item(0, 0).setCheckState(Qt.Unchecked)
+    assert not dialog.get_resolver().is_saved(Dependency("dataset", 1))
+
+
+def test_file_backed_dataset_is_not_a_checkable_root():
+    _qapp()
+    app_state = _FakeAppState([_FakeDataset(1, [], "ram"), _FakeDataset(2, ["/data/dem.tif"], "on-disk")])
+    dialog = SaveProjectDialog(app_state, None)
+
+    # Only the RAM-backed dataset gets a checkbox row...
+    assert dialog._roots_table.rowCount() == 1
+    # ...and the file-backed dataset is saved regardless (referenced, not a decision).
+    assert dialog.get_resolver().is_saved(Dependency("dataset", 2))
+
+
+def test_app_module_imports():
+    # Catches syntax/import errors from the app.py menu-wiring edits.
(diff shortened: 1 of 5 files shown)


Joooshe (changes requested):
Great job hooking things up. I do think though that we need a way for the user to see everything that is going to be saved. Or at least something that will be saved. It just feels like if I go to this screen and I have datasets (disk backed) loaded it shouldn't be empty

Image

Maybe we can add a third section with a tree widget that shows included datasets and ROIs and under the datasets it shows the things that depend on that dataset that are saved. And under the ROI it shows things that depend on the ROI that are saved. That way the user doesn't just see 2 blank sections. Instead they will see 2 blank sections and one populated section.

mmaclay:
Okay tried to add what I think you are talking about... does this make sense? One scenario is the result is checked, and one is unchecked:
save_dialog
save_dialog_excluded

PR #682, opened 2026-07-08
feat(project): project-file versioning policy + migration seam (#628)

### LINKED ISSUE #628: [Project File] Project versioning & migration

mmaclay:
## What does this change do?
Owns the project-file versioning policy and enforces its mechanism: documents the backward-compatibility guarantee and the additive-vs-breaking rule, adds a `register_migration` seam, and tests the migrate-up chain the future golden-file suite will rely on.

- [x] Feature
- [ ] Bug Fix
- [x] Documentation Update
- [ ] Style
- [ ] Code Refactor
- [x] Test
- [ ] Hot Fix
- [ ] Build
- [ ] CI
- [ ] Chore
- [ ] Revert

## Why is this change needed?
A project-file format that users save will change as WISER evolves. Without an explicit, enforced versioning strategy, either a schema change silently breaks old files, or the load path accretes ad-hoc branching. The bundle (#616) already carries a `format_version` and a migration *seam*, and the load orchestrator checks the version — but neither commits to a *policy* or a *mechanism* proven by tests. This issue owns that.

Closes #628. Part of epic #615. Builds on the `format_version` + migration seam from #616 and is invoked by the load orchestrator #627.

**Stacked on #626** (`feat/proj-626-save-dialog`, PR #681). Its diff shows only the versioning changes — review/merge the stack below it first, after which this rebases onto main.

- **`project/migrate.py`** — expands the module docstring into the versioning **policy**: the guarantee (*any file written by a released WISER opens in every later WISER*; forward compatibility is not promised), the **additive-vs-breaking rule** (a new *optional* field needs no bump and no migration — `from_pyrep` parses leniently and the load orchestrator reads sections with `.get`; bump the version and write a migration only for a renamed/removed/restructured/semantically-changed field), and the **version-bump checklist**. Adds `register_migration(from_version, fn)` for appending pure `migrate_vN_to_vN+1` transforms to the chain.
- The migrate-up mechanism and refuse-too-new behavior already existed in `migrate_up` / `ProjectBundle.read_manifest`; this PR proves them with tests and documents the policy around them.

**The golden-file regression fixture is intentionally deferred.** Per the epic's own guidance it's the most deferrable scope — there is only format v1 today and no migration to bridge, and the schema may still settle before the first release. The version-bump checklist in `migrate.py` is the trigger to capture a golden `.wiserproj` (written by the previous release) and add a load-it-on-current-code regression test at the first bump.

## Added tests?
- [x] yes

`test_project_versioning.py` (5): a registered migration chain applies step-by-step in order; `register_migration` wires into the chain; a missing chain step is refused (`ProjectFormatError`); a too-new bundle is refused cleanly (`ProjectTooNewError`) at the bundle level; and additively-added unknown keys survive migration untouched. The chain tests monkeypatch `_MIGRATIONS` / `CURRENT_FORMAT_VERSION` (auto-reverted, no leak). Full project suite green (78 passed); ruff + ruff-format clean.

## Added to documentation?
- [x] yes — the versioning policy + version-bump checklist live in the `migrate.py` module docstring where contributors will see them.

## Checklist
- [x] There is an issue associated with this pull request (#628)
- [ ] Branch is up-to-date with main/master — stacked on `feat/proj-626-save-dialog`; rebases onto main after the stack below it lands

- Merge order: the stack merges bottom-up; this is the top — after #681 lands, re-parent/rebase this onto main.
- Deferred: capture the first golden-file fixture + its regression test when `format_version` is first bumped (checklist in `migrate.py`). This is the final piece of epic #615 — once the stack merges, WISER can save and reopen a real `.wiserproj`.

### DIFF
--- a/src/wiser/project/migrate.py
+++ b/src/wiser/project/migrate.py
@@ -1,12 +1,34 @@
-"""Project-file format versioning and the migrate-up seam.
+"""Project-file format versioning, policy, and the migrate-up seam (issue #628).
 
-The loader only ever understands the *current* schema.  On load, an older
-manifest is transformed step-by-step up to the current shape by a chain of pure
-functions; a manifest newer than this WISER understands is refused cleanly.
+**Guarantee:** any project file written by a released WISER can be opened by
+every later WISER (backward compatibility).  Forward compatibility is *not*
+promised -- a file from a newer WISER is refused cleanly rather than
+mis-interpreted.
 
-This module provides the version constant and the migrate-up mechanism only.
-The migration *policy* (the per-version transform functions) and the golden-file
-regression suite that enforces "old files still load" are owned by issue #628.
+**How it works.** The loader only ever understands the *current* schema.  On
+load, an older manifest is transformed step-by-step up to the current shape by a
+chain of pure ``migrate_vN_to_vN+1`` functions registered here; a manifest newer
+than this WISER understands raises :class:`ProjectTooNewError`.  The rest of the
+load code (every persister) therefore only ever sees the current shape.
+
+**Additive vs. breaking (the rule contributors follow).** A *non-breaking*
+change -- adding a new optional field -- needs **no** version bump and **no**
+migration: every ``from_pyrep`` parses leniently (ignores unknown keys, defaults
+missing ones), and the load orchestrator reads manifest sections with ``.get``,
+so an unknown section is simply ignored.  Bump :data:`CURRENT_FORMAT_VERSION` and
+write a migration **only** for a *breaking* change -- a renamed, removed,
+restructured, or semantically-changed field.
+
+**On a version bump (checklist):**
+1. Write a pure ``migrate_v{N}_to_v{N+1}(manifest) -> manifest`` and
+   :func:`register_migration` it; unit-test it with a minimal before/after dict.
+2. Bump :data:`CURRENT_FORMAT_VERSION` to ``N+1``.
+3. Capture a **golden fixture** -- a real ``.wiserproj`` written by the *previous*
+   release -- and add a regression test that loads it on current code and asserts
+   a correct restore.  This is what converts the guarantee from aspiration into
+   something CI fails on.  (The golden-fixture suite is intentionally deferred
+   until the format stabilizes for the first release -- there is only v1 today and
+   no migration to bridge -- but this checklist is the trigger for adding it.)
 """
 
 from typing import Any, Callable, Dict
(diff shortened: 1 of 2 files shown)


src/wiser/project/migrate.py
```diff
@@ -23,11 +45,27 @@ class ProjectTooNewError(ProjectFormatError):
 
 
 # Maps a from-version to the function transforming a manifest dict to the next
-# version.  Empty while v1 is current; issue #628 appends migrations here as the
-# schema evolves (e.g. ``{1: migrate_v1_to_v2}``).
+# version.  Empty while v1 is current; migrations are appended via
+# :func:`register_migration` as the schema evolves (e.g. ``1 -> migrate_v1_to_v2``).
 _MIGRATIONS: Dict[int, Callable[[Dict[str, Any]], Dict[str, Any]]] = {}
```
Joooshe, line 51: Is the idea that the migrations would be registered in this file at module-level scope? If so maybe add a commented out example like 
```
"""
Example of registering a migration
register_migration(1, migrate_v1_to_v2)

def migrate_v1_to_v2(manifest: Dict[str, Any]) -> Dict[str, Any]:
   ...
"""
```

That way in the future when a migration does occur we know the protocol that we currently have in mind.

Joooshe: Also we should make sure the documentation that is in `save-to-project.md` is up to date.

Joooshe: Sorry, I didn't realize i forgot to submit these

PR #709, opened 2026-07-13
Unified Save-Project dialog: one tri-state selection tree

mmaclay:
## What does this change do?
Replaces the three-panel Save-Project dialog with a single checkable tree — three groups
(Datasets, ROIs, Analysis outputs), every item checked by default — that is both the
selection UI and a live consequence preview. Adds a "Save self-contained" portability
checkbox and Include/Exclude All, wires the chosen selection + mode into `save_project`, and
updates the developer guide.

- [x] Feature
- [x] Documentation Update
- [x] Test

## Why is this change needed?
Same manual-QA feedback as PR-A: the old dialog only exposed in-memory datasets across three
confusing panels and couldn't deselect a file-backed dataset or an individual analysis
output. This is the UI half, consuming PR-A's model.
Related: save-to-project epic (#615–#628). 

- `save_project_dialog.py` is rewritten as one checkable `QTreeWidget` built from
  `save_plan.save_tree`. Group parents are tri-state; toggling a group or dataset/ROI
  cascades to its products, whose `(snapshot)`/`(dropped)` consequence is annotated **in
  place** on the child. The tree is built once and updated in place on every toggle — never
  rebuilt — so the item whose `itemChanged` is being handled is never deleted, and
  cross-subtree cascades (excluding a dataset marks its ROI-average under the ROI node)
  update correctly.
- `app.py` passes the dialog's resolver + self-contained flag to `save_project` and remembers
  the mode for an in-place re-save.
- `save-to-project.md` documents both new features.

## Added tests?
- [x] yes — three-group structure; dataset/group/ROI exclusion reflected in the resolver;
  file-backed dataset now deselectable; Include/Exclude All; group tri-state
  (PartiallyChecked); in-place cascade re-annotation; self-contained checkbox; app-import
  smoke. 112 project tests green.

## Added to documentation?
- [x] yes — `save-to-project.md` (Overview save modes, resolver exclusion, a "Choosing what
  to save" tree section, the `save_project` signature + sequence diagram, and the extending
  contract).

## Checklist
- [ ] There is an issue associated with this pull request  
- [ ] Branch is up-to-date with main  
- [ ] All CI checks passed

save_dialog_default
save_dialog_excluded

Merges after PR-A; both merge with the epic before release.


Joooshe:
Quick first look. If there are no ROIs or analysis outputs, having the check box / section  just greyed out would be good. Looking at it right now so i will add more as i go.

Joooshe:
Maybe for when an ROI's average spectrum is calculated on two different datasets, we could have parentheses at the end saying the dataset it belongs to (or something to indicate). 
image

However, if you think this is not that useful to add, that is fine.

src/wiser/gui/app.py
```diff
@@ -686,7 +687,11 @@ def on_save_project(self, checked=False):
             self.on_save_project_as()
             return
         try:
-            save_project(self._app_state, self._current_project_path)
+            save_project(
```
Joooshe, line 696: Right now, if a user opens the save project dialog, does some changes, closes it, and re opens it, then the save project dialog does not keep any state. I think it probably should keep state, but let me know what you think. On that note, if it does keep state, then in this function `def on_save_project` we would need to get the resolver from this saved state so we know what to save based on what the user entered before they closed the dialog.

mmaclay: Good Catch! I fixed this in the last commit so the selection now lives with the project alongside its path and storage mode.

Joooshe:
Would be good to make save_project be on a separate thread with a progress bar. Use `src\wiser\utils\progress.py` and `src/wiser/gui/progress_task.py` but make it modal and non-cancellable. The function run_with_progress has been used to make these, but you may need to edit it or the other files for your use case.

PR #743, opened 2026-07-28
feat: Making dataset matching better

### LINKED ISSUE #667: WISER can't open up .hdr and .dat file combinations
## Describe the bug
As title says. WISER should be able to because if you remove the .dat then it can open it.

### LINKED ISSUE #744: Refactor raster format dispatch: replace try-every-loader with an identify-based registry
## Summary

`RasterDataLoader.load_from_file` picks a format by attempting every registered
loader against the file and keeping whichever one happened to succeed. This is
non-deterministic in a way that isn't obvious from reading it, opens (and leaks)
a GDAL handle per attempt, and runs an interactive dialog as part of probing
rather than after a format has been chosen.

### Format registry

Replace `_formats` and `_format_loaders` with one registry of specs:

```python
@dataclass(frozen=True)
class FormatSpec:
    name: str                     # "ENVI", "NetCDF", ... — the override token
    impl: Type[RasterDataImpl]
    extensions: frozenset[str]    # ordering hint, NOT an exclusion filter
    priority: int                 # explicit tiebreak, replaces dict order
    loader: Callable              # replaces the impl-type -> function dict
    interactive_step: bool        # declares that a GUI phase exists
```

Registering a format becomes one entry instead of three, which structurally
prevents the FITS/JP2-PDR drift described above.

Joooshe:
## What does this change do?

Makes dataset matching better. Closes #667 Closes #744

Rewrites how WISER decides *which* raster format should open a given file. Format
selection used to be an accident of dict ordering; it is now a declared,
inspectable registry with a cheap identification step and an explicit override.

> **Note on the branch name.** This branch is called `feat/ds-load-thread`, but
> **it does not add threading.** Nothing here moves work off the UI thread. What
> it does is restructure dataset loading so that developers have control
> over which format opens a file — which is the prerequisite that was missing
> before threading could be attempted.

- [ ] Feature
- [x] Bug Fix
- [x] Documentation Update
- [ ] Style
- [x] Code Refactor
- [x] Performance Improvements
- [x] Test
- [ ] Hot Fix
- [ ] Build
- [ ] CI
- [ ] Revert

## Why is this change needed?

`RasterDataLoader.load_from_file` picked a format by attempting **every**
registered loader against the file and keeping whichever happened to succeed.
The loop had no `break`, so `impl_list` was reassigned on every success and the
**last** format to succeed won — meaning which implementation opened a file
depended on the declaration order of a dict, not on anything about the file.

That had several consequences:

- **No control over format selection.** A developer could not say "open this as
  ENVI". There was no way to express intent, and no way to predict which format
  would be chosen for an ambiguous file.
- **No granularity per file type.** Every format got the same treatment. There
  was nowhere to hang behavior specific to one format.
- **Adding a format meant editing three separate structures** (`_formats`,
  `_format_loaders`, and the impl itself), which had already drifted apart.

### Bugs this fixes

| Bug | Detail |
|-----|--------|
| **Last-match-wins format selection** | Missing `break` in the attempt loop meant format priority was reverse dict-declaration order: PDS4 > PDS3 > JP2 > ASCII > NetCDF > GTiff > ENVI. An `.img` that both the ENVI and PDS drivers accepted resolved to PDS3. |
| **Leaked GDAL handles** | Every successful attempt opened a `gdal.Dataset` and all but one were discarded, so a file several drivers accepted was fully opened several times per load. |
| **ENVI `.dat` cubes could not be opened** | `get_load_filename` tried only the bare stem and `.img`, so selecting `scene.hdr` beside `scene.dat` raised `ValueError` — despite `*.hdr` being offered in the file-open dialog. |
| **GeoTIFF `.tiff` world files could not be resolved** | `.tfw` mapped only to `.tif`, never `.tiff`, though the dialog filter offers `*.tiff`. |
| **ENVI claimed unrelated rasters** | GDAL's ENVI driver identifies any file with a same-stem `.hdr` beside it, so a stray `scene.hdr` next to an unrelated `scene.tif` made the GeoTIFF look like an ENVI dataset. |
| **Projects re-detected format on every open** | A saved project recorded no format, so a registry change or a new file appearing beside the data could silently reopen a dataset through a different implementation than the one it was saved with. |
| **Platform-dependent test** | `test_subdataset_recorded_as_base_path_and_descriptor` hardcoded POSIX paths and failed on Windows (`'C:\data\scene.nc' == '/data/scene.nc'`). |

**A format registry.** `src/wiser/raster/format_registry.py` declares one
`FormatSpec` per format — implementation class, claimed extensions, priority,
dataset-construction strategy, and whether opening can block on a dialog.
Registering a format is now a single entry rather than three, so the impl and
the dispatch tables cannot drift apart.

**Extensions order, they do not gate.** Candidates are ranked
extension-claimants first (by priority), then everything else (by priority), then
the GDAL catch-all. A format is never *excluded* for failing to claim the path's
extension, because extensions here are genuinely ambiguous (`.img` is ENVI *and*
PDS3; `.hdr` is a raster header *and* a spectral library; `.xml` is PDS4 and much
else). This keeps files with unusual or absent extensions loadable.

**A cheap `identify()` before any open.** Each candidate is asked whether it
recognizes the file; the first positive match wins and the search stops. For
GDAL-backed formats this delegates to `gdal.IdentifyDriverEx`, which runs the
driver's own `Identify()` **without opening the dataset** — so WISER does not
duplicate GDAL's magic-byte checks and cannot drift from them. Only the winner
is opened: one file handle per load instead of one per accepting driver.

**Explicit overrides.** `load_from_file(..., format="ENVI")` skips detection
entirely and fails loudly rather than silently guessing again. Three callers use
it: programmatic callers, the file-open dialog (the user's selected filter was
previously discarded), and project restore.

**Declared sidecar resolution.** `DATA_EXTENSIONS` / `SIDECAR_EXTENSIONS` replace
the per-class `get_load_filename` implementations. The same list does double
duty: it resolves `scene.hdr` → `scene.dat`, *and* it stops a format claiming a
file whose extension it does not recognize.

### Files touched

| File | Change |
|------|--------|
| `src/wiser/raster/format_registry.py` | **New.** `FormatSpec`, `RASTER_FORMATS`, `candidates_for`, `get_format`, `format_for_impl`. |
| `src/wiser/raster/dataset_impl.py` | `Confidence` enum; `identify()` on the base and on `GDALRasterDataImpl`; generic `get_load_filename`; driver declarations per subclass. |
| `src/wiser/raster/loader.py` | `load_from_file` rewritten around order → identify → open-the-winner; `format=` parameter. |
| `src/wiser/raster/dataset.py` | `get_impl()` accessor. |
| `src/wiser/gui/app.py`, `app_state.py` | Dialog filters paired with format names; selected filter threaded through as an override. |
| `src/wiser/project/persisters/datasets.py` | Record the opening format in the manifest; honor it on restore. |

## Added tests?
- [x] yes
- [ ] no, because they aren’t needed
- [ ] no, because I need help

26 new tests in `src/tests/test_format_registry.py` covering registry shape,
candidate ordering, `identify()` behavior (including that it opens no handle),
the single-open guarantee, override semantics, and both sidecar regressions.
3 new tests in `src/tests/test_project_datasets.py` for the manifest format field
— round-trip, a manifest written before the field existed, and one naming a
format no longer registered.

## Added to documentation?
- [x] yes
- [ ] no documentation needed

- `developer-content/raster-format-dispatch.md` — **new.** How dispatch works,
  why extensions order rather than gate, and a step-by-step guide with a
  checklist for adding a new file type.
- `user-content/opening-data-files.md` — **new.** Explains that the file type
  chosen in the Open dialog matters, which file to pick for multi-file datasets,
  and troubleshooting.
- `index.rst` — format list corrected (FITS and ASCII Grid were missing; PDS
  split into PDS3 and PDS4).

## Checklist
- [X] There is an issue associated with this pull request
- [X] No new lint/style issues introduced
- [X] Branch is up-to-date with main/master
- [ ] All CI checks passed

Not applicable — no visible UI change beyond the file-open dialog's type list,
which gains **FITS** and **ASCII Grid** entries and splits the single **PDS**
entry into **PDS3** and **PDS4** (a combined PDS filter cannot name one format,
so the override would have been inert for it).

1. **Confirm the ENVI data-extension list against real data.** It is currently
   `("", ".img", ".dat")`. `.bin`, `.bsq`, `.bil`, and `.bip` also appear in the
   wild and may be worth adding.

2. **Developer environment note (not a merge blocker).** If netCDF tests fail
   locally with `Can't load requested DLL: ...gdal_netCDF.dll`, check for a
   machine-scope `GDAL_DRIVER_PATH` pointing at a *different* conda environment.
   Conda envs set this themselves on activation; a stale machine-level value
   overrides them and loads plugins from the wrong GDAL build.

### DIFF
--- a/doc/sphinx-general-wiser-docs/source/developer-content/code-documentation.md
+++ b/doc/sphinx-general-wiser-docs/source/developer-content/code-documentation.md
@@ -23,4 +23,5 @@ crs-creator-internals.md
 plugin-system.md
 mosaic-internals.md
 save-to-project.md
+raster-format-dispatch.md
 ```
--- a/doc/sphinx-general-wiser-docs/source/developer-content/raster-format-dispatch.md
+++ b/doc/sphinx-general-wiser-docs/source/developer-content/raster-format-dispatch.md
@@ -1,0 +1,352 @@
+# Raster Format Dispatch
+
+When WISER is handed a file path, something has to decide *which* of the raster
+formats it knows about should open it. That decision is **format dispatch**, and
+it lives in `src/wiser/raster/format_registry.py` and
+`RasterDataLoader.load_from_file`.
+
+This page documents how a path becomes a `RasterDataSet`, and — the part most
+developers come here for — [how to add support for a new file
+type](#adding-a-new-format).
+
+---
+
+## Overview
+
+Dispatch is deliberately split into three stages, so that the cheap decisions
+happen before the expensive ones and nothing irreversible happens while WISER is
+still guessing.
+
+| Stage | What it does | Cost |
+|-------|--------------|------|
+| **Order** | Rank the registered formats for this path, most likely first. | String comparison only. |
+| **Identify** | Ask each candidate, in order, whether it recognizes the file. | A few bytes read per candidate. No handles opened. |
+| **Open** | Open the file with the format that won. | One file handle, once. |
+
+```{mermaid}
+flowchart TB
+    Path["load_from_file(path)"]
+    Override{"format= given?"}
+    One["Use that format only.<br/>Fail loudly if it doesn't work."]
+    Order["candidates_for(path)<br/>extension-claimants first, then<br/>everything else, then the catch-all"]
+    Ident["identify() down the list"]
+    Yes{"Confidence.YES?"}
+    Win["Winner"]
+    Maybe["Highest-priority MAYBE"]
+    Open["try_load_file() on the winner"]
+    Build["FormatSpec.loader ->  RasterDataSet(s)"]
+
+    Path --> Override
+    Override -- yes --> One --> Open
+    Override -- no --> Order --> Ident --> Yes
+    Yes -- "stop immediately" --> Win --> Open
+    Yes -- "none were certain" --> Maybe --> Open
+    Open --> Build
+```
+
+### Why not just try every format?
+
+That is what WISER used to do, and it caused three problems worth remembering,
+because each one is a rule the current design exists to enforce:
+
+- **The winner was whichever format was tried last.** The old loop had no
+  `break`, so every format was attempted and the last success overwrote the
+  earlier ones. Which implementation opened a file therefore depended on the
+  declaration order of a dict.
+- **Every successful attempt opened a file handle**, and all but one were
+  discarded.
+- **Identification could block on the user.** The NetCDF sub-dataset dialog ran
+  inside the attempt, so it could be shown for a file that a later format then
+  claimed, throwing away what the user chose.
+
+Hence: exactly one format wins, `identify()` never opens or prompts, and only
+the winner is opened.
+
+---
+
+## The registry
+
+Every format WISER can open is one `FormatSpec` entry in `RASTER_FORMATS`:
+
+```python
+FormatSpec(
+    name="ENVI",                                    # the override token
+    impl=ENVI_GDALRasterDataImpl,                   # class that opens it
+    extensions=frozenset({"", ".hdr", ".img", ".dat"}),
+    priority=70,
+    loader=load_normal_dataset,
+    interactive_step=False,
+)
+```
+
+| Field | Meaning |
+|-------|---------|
+| `name` | Stable identifier. Used by the `format=` override and **written into project files**, so renaming one is a breaking change. |
+| `impl` | The `RasterDataImpl` subclass that opens this format. |
+| `extensions` | Extensions this format claims. An **ordering hint only** (see below). |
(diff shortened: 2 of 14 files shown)


Joooshe:
I think Ill just leave this one open for you @mmaclay . It should be ready to merge in, feel free to make changes or reject.

Joooshe:
Also I started [this ](https://github.com/Ehlmann-research-group/WISER/actions/runs/30417329848) run to ensure it works.

PR #778, opened 2026-08-28
fix: Store cache entries on a miss so the render and computation caches serve hits

### LINKED ISSUE #772: Cache.add_cache_item never stores new keys — the render and computation caches serve no hits
## Describe the bug

`Cache.add_cache_item` (`src/wiser/raster/data_cache.py:65`) guards its insert with `if key in
self._cache:` and returns `False` otherwise:

```python
def add_cache_item(self, key, value) -> bool:
    if key in self._cache:        # only updates keys that already exist
        ...
        self._cache[key] = value
        self._size += value.nbytes
        return True
    return False                  # a brand-new key is never stored
```

Every caller adds an item only **after** a cache miss, so the key is never already present.
Confirmed at the call sites in `src/wiser/raster/dataset.py` — `get_band_data`,
`get_band_data_normalized`, `get_image_data`, and `get_image_data_subset` all follow
`get_cache_item()` → `if arr is None:` → read → `add_cache_item()`.

Net effect: **the render cache and the computation cache never serve a hit.** Every render reads,
re-masks, and re-normalizes from scratch. `HistogramCache` overrides `add_cache_item` without the
guard, so the histogram cache works as intended.

## To Reproduce

1. Open a dataset.
2. Switch to another dataset and back, or drag a slider in the stretch builder.
3. The band data is re-read and re-normalized every time; `Cache._cache` stays empty for the
   render and computation tiers.

## Expected behavior

A miss followed by `add_cache_item` stores the entry, and the next equivalent request hits.

## Additional context

The condition is almost certainly meant to be `if key not in self._cache:`. Note the fix is not
purely a one-character change — the branch does `self._size += value.nbytes` unconditionally,
which double-counts on a genuine replacement. The insert path must add the size, and the
replacement path must subtract the old value's size first (or return early when the key is
already present with the same value).

**This is already written up as a known issue in the developer docs**
(`doc/sphinx-general-wiser-docs/source/developer-content/data-caching.md`, "Known Issue:
`Cache.add_cache_item` Guard") but was never filed as an issue.

**Sequencing note — this matters for #758.** Profiling the render path today measures a pipeline
in which the computation and render caches never hit, so the profile will attribute to the
rendering code cost that caching was designed to absorb. Fix this first, or profile both
configurations, before drawing conclusions about where the render time goes.

It also bears on #762: the stated value of `Dataset._data_cache` is that it avoids
re-normalization on the rendering hot path, and that benefit is not currently realised at all.

Related: #758, #762, #574 (rendering gets stuck — "this may be a cache problem"),
`doc/.../developer-content/data-caching.md`.

eastagiletracker:
This PR proposes fixing `Cache.add_cache_item` so that a miss actually stores its entry, which is what the render and computation caches need before they can serve a hit at all (Fixes #772). We include this PR work along with a full history of your repo at https://eastagiletracker.com/projects/454. You can sign in with your GitHub ID to claim ownership of the project.

## What is broken

`Cache.add_cache_item` guards its insert with `if key in self._cache:` and returns `False` otherwise, so a key that is not already present is never stored. Every caller reaches it only *after* a lookup missed — `get_band_data`, `get_band_data_normalized`, `get_image_data` and `get_image_data_subset` in `src/wiser/raster/dataset.py`, and `RasterView.update_display_image` in `src/wiser/gui/rasterview.py`, all follow `get_cache_item()` → `if arr is None:` → read → `add_cache_item()`. The key is therefore never already there, and the render and computation tiers stay empty for the life of the process. `HistogramCache` overrides `add_cache_item` without the guard, which is why the histogram tier is unaffected.

Reproduced on `main` at `d0b14c4b`, loading one of the repo's own test datasets through `RasterDataLoader` with a `DataCache` attached and calling `get_band_data_normalized(0)` 51 times:

```
[file] test_utils/test_datasets/caltech_4_100_150_nm
[file] first read        : 2360 us
[file] repeat read (avg) : 1228 us over 50 calls
[file] computation cache : 0 entries, 0 bytes
[cube] shape (16, 700, 700) float32
[cube] repeat read (avg) : 177 us over 50 calls
[cube] computation cache : 0 entries, 0 bytes
```

Nothing is held, so every one of those reads goes back to GDAL and re-normalizes — the work `_data_cache` exists to absorb on the rendering hot path.

## The change

`src/wiser/raster/data_cache.py`, all inside the base `Cache`:

- The capacity check moves up front and the insert becomes unconditional, so a miss stores its entry.
- A replacement gives back the old value's `nbytes` before adding the new value's, so `_size` counts an overwritten key once rather than twice. Deleting before re-inserting also moves the refreshed entry to the back of the `OrderedDict`, so FIFO eviction treats it as the newest item rather than the oldest.
- `_evict()` now runs *after* the insert, once `_size` includes the new item, so it frees enough room for the value being added. With the old ordering the loop compared a size that did not yet count the incoming array and could leave the cache above its capacity — which never showed up because nothing was ever stored. These are 10 GB tiers of raster arrays, so that mattered as soon as the insert started working.
- `clear_cache()` needed one more line to survive the fix: it deleted from the `OrderedDict` it was iterating, which was harmless only while the cache was always empty. On a populated cache it raises `RuntimeError: OrderedDict mutated during iteration`, so it now iterates a snapshot of the keys.

`HistogramCache.clear_cache` has the same iterate-and-delete shape and is reachable today, since that tier already stores. I left it out to keep this to the one defect — happy to send it separately if you want it.

The developer docs carried this as a known issue, so `doc/.../developer-content/data-caching.md` swaps that warning for a short description of the store-on-miss behaviour and the size accounting that goes with it.

## Verification

Same measurement on the patched tree, unchanged script:

```
[file] test_utils/test_datasets/caltech_4_100_150_nm
[file] first read        : 1961 us
[file] repeat read (avg) : 1 us over 50 calls
[file] computation cache : 1 entries, 90000 bytes
[cube] shape (16, 700, 700) float32
[cube] repeat read (avg) : 1 us over 50 calls
[cube] computation cache : 1 entries, 1960000 bytes
```

`src/tests/test_data_cache.py` is new and covers the miss/store/hit cycle for both tiers, masked arrays from the data-ignore path, the size arithmetic across replacement, removal and `clear_keys_from_partial`, and FIFO eviction at capacity. Ten of its fifteen cases fail against unmodified `main` and all fifteen pass with the change; the five that pass either way are the controls — an over-capacity value is still refused, and `HistogramCache` still round-trips its `(bins, edges)` tuple.

```
$ cd src/tests && pytest test_data_cache.py -q          # on unmodified main
10 failed, 5 passed in 3.40s
$ cd src/tests && pytest test_data_cache.py -q          # with this change
15 passed in 3.10s
```

The rest of the suite was run the way `dev-CI.yml` runs it — under `xvfb-run -a --server-args="-screen 0 1280x1024x24 -ac -nolisten tcp"`, in the `linux-cached-dev` environment built from `etc/dev-conda-lock.yml` — and is entirely green with this change: 890 passed, 5 skipped, no failures and no errors across all 91 existing files in `src/tests`. `ruff format --check .` reports 300 files already formatted and `ruff check .` is clean, both on `ruff==0.1.15` as CI pins it.

## How this was managed

This work was tracked on a board imported from this repository's issues and pull requests, on the story for this defect: https://eastagiletracker.com/projects/454/stories/400865 — the board itself is at https://eastagiletracker.com/projects/454, holding 605 stories brought over from your issues, pull requests and milestones.

board

If you'd rather not receive contributions like this, reply `no-more-prs` on this pull request and we won't open any further ones on your repositories.

---

**Lawrence W. Sinclair**
CEO / East Agile
[linkedin.com/in/lwsinclair/](https://linkedin.com/in/lwsinclair/)
[eastagile.com](https://eastagile.com)

### DIFF
--- a/doc/sphinx-general-wiser-docs/source/developer-content/data-caching.md
+++ b/doc/sphinx-general-wiser-docs/source/developer-content/data-caching.md
@@ -4,8 +4,8 @@ Rendering a raster is expensive: reading and normalizing bands, applying
 stretches, and packing pixels all cost real time, and the same work is requested
 repeatedly as the user scrolls, switches panes, or reopens dialogs. WISER avoids
 recomputation with a three-tier in-memory cache. This page documents what each
-tier stores, how keys and eviction work, the cache lifecycle, and a known issue
-to be aware of.
+tier stores, how keys and eviction work, the cache lifecycle, and how storing on
+a miss keeps the running size and eviction in step.
 
 This is the caching half of the [Rendering Pipeline](rendering-pipeline.md).
 
@@ -186,33 +186,26 @@ flowchart TD
 
 ---
 
-## Known Issue: `Cache.add_cache_item` Guard
+## Storing on a Miss
 
-```{warning}
-The base `Cache.add_cache_item`
-(`src/wiser/raster/data_cache.py`) currently guards its insert
-with `if key in self._cache:` and returns `False` otherwise:
+`Cache.add_cache_item` stores the entry whether or not the key is already
+present. That is what lets the render and computation caches serve hits at all:
+callers only reach it *after* a lookup missed, so the key is never already there
+and an insert-only-if-present guard would drop every new entry.
 
-​```python
-def add_cache_item(self, key, value) -> bool:
-    if key in self._cache:        # <-- only updates EXISTING keys
-        ...
-        self._cache[key] = value
-        self._size += value.nbytes
-        return True
-    return False                  # <-- a brand-new key is never stored
-​```
+Two details keep `_size` honest, and eviction with it:
 
-Callers add an item only **after** a cache miss (`in_cache(key)` returned
-`False`), so the key is never already present — which means new entries are
-never stored. As written, the **render cache and computation cache do not
-actually serve hits**: every render recomputes from scratch. The condition is
-almost certainly meant to be `if key not in self._cache:`.
+- **A replacement subtracts before it adds.** Re-storing a key that is already
+  in the cache gives back the old value's `nbytes` before adding the new
+  value's, so an overwritten key is counted once rather than twice. The
+  refreshed entry moves to the back of the `OrderedDict`, so FIFO eviction
+  treats it as the newest item rather than the oldest.
+- **Eviction runs after the insert.** `_evict()` is called once the new item is
+  in and `_size` includes it, so the loop frees enough room for the incoming
+  value instead of measuring a size that does not yet count it. A value larger
+  than the entire capacity is refused up front — `add_cache_item` returns
+  `False` and nothing is evicted on its behalf.
 
-`HistogramCache` overrides `add_cache_item` *without* this guard, so the
-histogram cache works as intended.
-
-This page documents the *intended* design (the caches are meant to store on
-miss and serve on subsequent requests). The note is here so readers aren't
-misled by the current code; fixing the guard is a separate change.
-```
+`src/tests/test_data_cache.py` covers the miss/store/hit cycle for the render
+and computation caches, the size arithmetic across replacement and removal, and
+FIFO eviction at capacity.
--- a/src/tests/test_data_cache.py
+++ b/src/tests/test_data_cache.py
@@ -1,0 +1,239 @@
+"""Unit tests for the three-tier in-memory cache in ``wiser.raster.data_cache`` (#772).
+
+Small NumPy arrays only -- no widget, no ``QApplication``, no GDAL. These cover the
+contract every caller of the render and computation caches relies on: a miss followed
+by ``add_cache_item`` stores the entry, the next equivalent request hits, and the
+running byte count stays in step with what is actually held so that eviction keeps the
+cache inside its capacity.
+"""
+import unittest
+
+import numpy as np
+import pytest
+
+import tests.context  # noqa: F401 -- adds src/ to sys.path
+
+from wiser.raster.data_cache import (
+    Cache,
+    ComputationCache,
+    DataCache,
+    HistogramCache,
+    RenderCache,
+)
+
+pytestmark = [
+    pytest.mark.smoke,
+    pytest.mark.unit,
+    pytest.mark.regression,
+]
+
+# The caches only need the "dataset" to be hashable, so a sentinel string stands in for
+# a RasterDataSet throughout and keeps these tests free of GDAL.
+_DATASET = "dataset-a"
+_OTHER_DATASET = "dataset-b"
+
+
+def _array(n_bytes: int) -> np.ndarray:
+    """
+    Returns a float64 array occupying exactly n_bytes bytes.
+
+    Args:
+        n_bytes (int): The size of the array in bytes; must be a multiple of 8.
+    """
+    assert n_bytes % 8 == 0, "float64 arrays are sized in multiples of 8 bytes"
+    return np.arange(n_bytes // 8, dtype=np.float64)
+
+
+class TestAddCacheItemStoresNewKeys(unittest.TestCase):
+    """A miss followed by an add must be served from the cache next time."""
+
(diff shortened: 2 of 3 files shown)


mmaclay:
Thanks for this — the diagnosis is correct, and the three fixes inside `Cache` are the right shape: unconditional insert, subtract-before-add on replacement, and `_evict()` after the insert rather than before. I reproduced your benchmark and your suite run (890 passed, 5 skipped on the patched tree), so the numbers in the description hold up.

I'm closing this rather than merging, because the change isn't a guard fix. `add_cache_item` has been refusing every insert since it was written, which means the render and computation tiers have never held anything — and several latent bugs behind them land the moment they do. I reproduced four against this branch.

**1. `get_image_data_subset` shares the whole-cube key.** `dataset.py:1015` builds `cache.get_cache_key(self)` — byte-identical to `get_image_data`'s key at `dataset.py:978`. The comment says "if it is we just splice", but there is no splicing code. Once the cube is cached, every subset request returns the entire cube:

```
subset before full-cube read : shape=(2, 4, 5)   expected (2, 4, 5)
subset after  full-cube read : shape=(6, 20, 30) expected (2, 4, 5)
```

That is aimed straight at the chunked-read path — `primitives.py:219` and `storage_client.py:569` both feed this a `DatasetRegionRef` and transpose the result expecting `(dy, dx, dband)`. The worker side escapes only because `storage_client.py:731` reconstructs datasets with `data_cache=None`. `continuum_removal_plugin.py:416` does not escape.

**2. `filter_data_ignore_value` is not in the key.** There's a standing `TODO` at `dataset.py:987` saying so. Inert today; not inert here. An unmasked read caching first is the common ordering — `decorrelation_stretch.py:425` and most task stages pass `filter_data_ignore_value=False` — and a later caller that explicitly asks for masking gets raw sentinels back with no error:

```
requested masking -> MaskedArray=False, sentinel -9999.0 returned as data
band mean = -617.4   # correct masked mean excludes -9999
```

**3. Closing a dataset leaks its cached bands.** `app_state.py:455-458` purges the computation cache with a single key — the whole-cube one. Every per-band entry survives and is unreachable forever:

```
after reading 4 bands raw+normalized      : 8 entries
after remove_dataset()'s single-key purge : 8 entries
```

**4. The capacities are invisible to the rest of WISER's memory model.** `DataCache` defaults to 10 GB render + 10 GB computation (`data_cache.py:305-306`) against `SCHEDULER_RAM_BUDGET = 2_000_000_000` (`work_scheduler.py:36`), which the scheduler treats as the whole budget and cannot see past. `bandmath/utils.py:911` sizes chunks from `psutil.virtual_memory().available` at a single instant, with no way to ask the cache to release.

Separately, cached arrays are handed out by reference and `continuum_removal_plugin.py:429` writes NaN through one, which poisons the entry for the next reader. Someone anticipated this — `rasterview.py:881` carries "it is held by the cache for subsequent renders" — but the invariant is currently unenforceable.

The green suite is the strongest argument for holding. All of the above are live in that run and no test noticed; there is no coverage of these paths, which is how the guard stayed broken long enough to get written up as a known issue.

One correction: `Cache.clear_cache` has no callers anywhere in `src/`, so `HistogramCache.clear_cache` isn't reachable today either.

I've rescoped #772 and filed the prerequisites as sub-issues — #784, #785, #786, #787, #788, #789. Your `Cache` change is correct and will land essentially as written once the keys and the lifecycle are fixed underneath it.

mmaclay:
no-more-prs

To be clear about why, since the diagnosis here was genuinely correct: WISER is a one-to-two developer project with a lot of stateful machinery, and the review cost of vetting an unsolicited PR against it is high — this one needed a full-suite run plus several reproductions to establish that a correct three-line fix was not safe to merge. That's not a reasonable ongoing load for us to carry.

Please don't open further pull requests on our repositories. The issue itself (#772) stays open and we'll take it from here.

PR #793, opened 2026-09-02
fix: apply netCDF scale_factor and add_offset on read

### LINKED ISSUE #791: netCDF: scale_factor / add_offset are ignored, so packed products read as raw integer counts
## Summary

WISER never applies a netCDF variable's `scale_factor` / `add_offset`. Products that store integer counts plus a linear transform open with the stored counts presented as physical values. Nothing warns, and the numbers are plausible enough to be mistaken for data.

## Steps to reproduce

1. Download a PACE OCI L2 AOP granule from [NASA Earthdata](https://search.earthdata.nasa.gov/) (verified on `PACE_OCI.20241121T193951.L2.OC_AOP.V3_2.nc`).
2. Open it and choose the `geophysical_data/Rrs` subdataset.
3. Click any water pixel and read the spectrum.

## What happens

Remote sensing reflectance comes back in the tens of thousands, negative. Every downstream tool inherits it: band-math indices, the stretch, statistics, classification, unmixing.

## Why

The variable is `Int16` and carries the transform as attributes. GDAL already parses them and exposes them per band:

| | |
|---|---|
| `GetScale()` | 2e-06 |
| `GetOffset()` | 0.05 |
| `GetNoDataValue()` | -32767 |
| Stored range | -32767 … -10933 |
| `raw × scale + offset` | **-0.0098 … 0.0281 sr⁻¹** |

The second row is correct Rrs. WISER calls neither `GetScale()` nor `GetOffset()` anywhere — a search for both across `src/wiser/` returns only the GUI's unrelated zoom `_scale_factor`.

## The part that makes this more than a missing multiply

The fill value is stored in the same integer space. Scaling the data without also moving the data-ignore value turns -32767 into **-0.015534**, which sits inside the physical range of the variable. Fill pixels would stop matching the mask in `RasterDataSet.get_band_data`, which compares against the raw sentinel, and would pass silently as ordinary measurements. Whatever applies the scaling has to carry the ignore value with it.

## Scope

This is not PACE-specific. Packing with `scale_factor` / `add_offset` is standard CF practice, so most operational L2 ocean, atmosphere and land products are affected. Unpacked netCDF is unaffected: EMIT L2A reflectance is `Float32` with no scale attributes, and GDAL reports `GetScale() = None` for it.

## Suggested fix

`NetCDF_GDALRasterDataImpl` already overrides all seven read paths (`get_image_data`, `get_image_data_subset`, `get_band_data`, `sample_band_data`, `get_all_bands_at`, `get_all_bands_at_rect`, `get_multiple_band_data`) to wrap them in `_quiet_gdal_warnings()`. Applying the transform in those wrappers, and overriding `read_data_ignore_value()` to report the scaled fill, confines the change to netCDF and leaves every other format untouched.

Since `scale_factor` / `add_offset` are attributes of a *variable*, all bands of one subdataset share them; bands that disagree mean the file does not match the netCDF data model and no single physical ignore value can describe it.

Found while writing a PACE lab for the tutorial documentation. Blocks that lab.

### LINKED ISSUE #792: NetCDF_GDALRasterDataImpl.get_multiple_band_data calls the wrong parent method and raises TypeError
## Summary

`NetCDF_GDALRasterDataImpl.get_multiple_band_data` delegates to `super().get_all_bands_at()` instead of `super().get_multiple_band_data()`. The signatures do not match, so the call raises `TypeError`.

## Where

`src/wiser/raster/dataset_impl.py`, in the netCDF override:

```python
def get_multiple_band_data(self, band_list_orig: List[int]) -> np.ndarray:
    with self._quiet_gdal_warnings():
        arr = super().get_all_bands_at(band_list_orig)   # <-- wrong method
    return arr
```

The parent signatures are:

```
GDALRasterDataImpl.get_all_bands_at(self, x, y)
GDALRasterDataImpl.get_multiple_band_data(self, band_list_orig: List[int]) -> np.ndarray
```

One positional argument against `(self, x, y)` gives:

```
TypeError: get_all_bands_at() missing 1 required positional argument: 'y'
```

The five neighbouring overrides in the same class each delegate to their own parent correctly; this one looks like a copy-paste that was never exercised.

## What triggers it

`RasterDataSet.get_multiple_band_data` is reached from:

- `src/wiser/bandmath/types.py` — when an expression binds a whole **image cube** variable rather than individual bands
- `src/wiser/gui/similarity_transform_dialog.py` (two call sites)
- `src/wiser/raster/georef_warp.py`

So: band math over a cube variable, or georeferencing, on any netCDF dataset. Band math that binds individual bands takes `IMAGE_BAND` → `get_data()` and does not hit it, which is why this has gone unnoticed.

## Suggested fix

```python
arr = super().get_multiple_band_data(band_list_orig)
```

Found while working on #791; both live in the same class and are fixed together in the accompanying PR.

mmaclay:
## What does this change do?

Packed netCDF variables store integer counts alongside the linear transform that turns them into physical units. WISER read the counts and presented them as data. This applies the transform on every netCDF read path, moves the data-ignore value and the declared element type into the same units, and refuses at open the packings where a scaled fill could no longer be told apart from a real measurement.

Also fixes a wrong `super()` call in the same class that made any netCDF multi-band read raise `TypeError`.

Closes #791. Closes #792.

- [ ] Feature
- [x] Bug Fix
- [ ] Documentation Update
- [ ] Style
- [ ] Code Refactor
- [x] Test
- [ ] Hot Fix
- [ ] Build
- [ ] CI
- [ ] Revert

## Why is this change needed?

PACE OCI L2 AOP stores `Rrs` as `Int16` with `scale_factor = 2e-06` and `add_offset = 0.05`. GDAL parses both and exposes them per band, but WISER called neither `GetScale()` nor `GetOffset()` anywhere. Opening the `Rrs` cube gave reflectance in the tens of thousands, negative, with no warning, and every downstream tool inherited it.

Verified on `PACE_OCI.20241121T193951.L2.OC_AOP.V3_2.nc`:

| | before | after |
|---|---|---|
| band 60 range | -32767 … -10933 (`Int16`) | **-0.0100 … 0.0285 sr⁻¹** (`float32`) |
| data-ignore value | -32767 | -0.015534 |
| declared element type | `int16` | `float32` |
| multi-band read | `TypeError` | works |

Regression check on EMIT L2A reflectance, which is unpacked `Float32`: unchanged in every respect, ignore value still -9999. GDAL reports `GetScale() = None` there, which short-circuits the whole path.

This is not PACE-specific. Packing with `scale_factor` / `add_offset` is standard CF practice, so most operational L2 ocean, atmosphere and land products are affected. One gap remains after this change: a netCDF holding a single data variable reports no subdatasets and falls back to the plain `GDALRasterDataImpl`, which does not scale. Tracked in #794.

The read side lives in `NetCDF_GDALRasterDataImpl`, which already overrode all seven read methods to wrap them in `_quiet_gdal_warnings()`. No other format is touched. Two call sites outside the class are, because they reach past those methods into the backing GDAL dataset — the last bullet.

- **`_read_scaling()`** reads the per-band scale/offset once at open and returns `None` when the variable is unscaled, which is the common case and leaves the read path as it was. `scale_factor` / `add_offset` are attributes of a *variable*, so a subdataset's bands all carry the same pair; bands that disagree raise, reporting the pair per band in band order. It does not sort them — a band reporting `None` where another reports a float is not orderable, and sorting would raise `TypeError` in place of the intended `ValueError`.

- **`_apply_scaling()`** applies the transform, and both the data and the data-ignore value go through it. That is load-bearing rather than tidy. `RasterDataSet.get_band_data` masks pixels against the reported ignore value *after* the impl returns, so scaling the data without moving the sentinel turns -32767 into -0.015534 — a value inside the physical range of `Rrs`. Fill pixels would stop matching the mask and pass silently as measurements. Both come out of the same arithmetic so they still compare equal.

- **`_unscaled_dtype()`** takes the output width from the stored width. Integers of 16 bits or fewer — the `int16` these products pack with — go to `float32`, which represents every such count exactly at half the memory of `float64` (1.5 GB rather than 3.0 GB for this granule). Wider integers go to `float64`, because `int32` counts above 2²⁴ are not exactly representable in `float32` and narrowing them would drop precision with nothing reporting it. A stored float keeps its own width: scaling cannot recover precision it never had.

- **`get_elem_type()`** reports that dtype rather than the stored integer one. This is the part that would have destroyed data. The export paths — `save_dataset_as`, `mosaic_materialize`, `similarity_transform_dialog` — ask for the element type and cast the read array to it, so a dataset still claiming `int16` writes a cube of zeros: every physical reflectance is under 0.5 and truncates, with `data ignore value = -0.015534` recorded in the header beside them.

- **`_validate_fill_separation()`** raises at open when the packing is too fine for the fill to stay distinguishable. The masking is `np.ma.masked_values`, which compares within `rtol=1e-5, atol=1e-8` rather than exactly, so whether a neighboring count falls inside that band depends on `offset/scale`. PACE is clear by about 12x. It inverts once `|offset/scale|` passes ~1e5: at `scale=1e-4, offset=100` the band spans roughly nine counts either side of the fill. Nothing we read today hits it, and packing with a large offset is common enough — brightness temperature, radiance — that failing at open beats dropping pixels quietly later.

- **`reads_are_transformed()`** is the guard for consumers that reach past the read methods. `georef_warp.warp_dataset_to_path` and `similarity_transform_dialog.create_translated_dataset` both hand the backing `gdal_dataset` to GDAL to stream, along with the dataset's data-ignore value. Once that value is physical and the backing dataset still holds counts, the two disagree: GDAL rounds `-0.015534` to `0` on an `Int16` VRT, so genuine zero-count pixels get dropped as fill while the real fill at -32767 is treated as valid and resampled into its neighbors. Both were self-consistent before this change. Scaled impls now take the per-band array paths, which read through the scaling.

The alternative shape, considered and not taken here: build the subdataset as a VRT with `gdal.Translate(..., unscale=True, outputType=..., noData=scaled_fill)` at open. `gdal_dataset` would then be physical for all seven read methods, `get_elem_type`, `GetNoDataValue` and every direct consumer at once, and `_unscale`, `get_elem_type`, `read_data_ignore_value` and both guards would go away. It is the better design, and `reads_are_transformed()` is an honor-system contract in the meantime — nothing stops the next `impl.gdal_dataset` call site from skipping the check.

It is not this PR because it breaks project save as written. `dataset.py:1759` takes `recreation_value = impl.gdal_dataset.GetFileList()[0]` for netCDF impls specifically, and against a `/vsimem` VRT that returns a path which does not survive the session. The VRT also has to outlive the impl across `reopen_dataset()`. Tracked in #795.

## Added tests?
- [x] yes
- [ ] no, because they aren't needed
- [ ] no, because I need help

`src/tests/test_netcdf_scaling.py`, 26 tests over a synthetic netCDF carrying a scaled `int16` variable and an unscaled `float32` one.

All seven wrapped read paths are covered — `get_image_data`, `get_image_data_subset`, `get_band_data`, `get_multiple_band_data`, `sample_band_data`, `get_all_bands_at` and `get_all_bands_at_rect` — along with: the fill stays masked and real pixels do not, the ignore value is reported in physical units, `get_elem_type()` follows the reads, an ENVI save-and-reload round trip returns physical values, and an unscaled variable is untouched in both value and dtype.

A second class drives the open-time checks without a file on disk: uniform, identity, absent and disagreeing scaling, including the unorderable `None`-against-float case; the fill-separation check accepting PACE's packing and rejecting a fine scale under a large offset; the dtype choice for narrow ints, wide ints and stored floats; and that the scaled fill still compares equal to the scaled data.

I confirmed these catch the defect: with `dataset_impl.py` reverted to `main`, 23 of the 26 fail.

Writing them turned up that GDAL presents netCDF rows bottom-up, so the array WISER returns is the stored band flipped in y. Noted in the test rather than worked around.

## Added to documentation?
- [ ] yes
- [x] no documentation needed

## Checklist
- [x] There is an issue associated with this pull request
- [x] Branch is up-to-date with main/master
- [x] All CI checks passed

Full suite locally: **921 passed, 5 skipped, 0 failed** (`QT_QPA_PLATFORM=offscreen pytest -q src/tests`, 34 min). All five CI checks green on `fbcd3f82`.

n/a — the visible effect is in the PACE lab figures, which are on the docs branch.

This unblocks the PACE lab in the tutorial documentation (PR #777), whose figures were captured against this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/tests/test_netcdf_scaling.py
+++ b/src/tests/test_netcdf_scaling.py
@@ -1,0 +1,163 @@
+"""Regression tests: netCDF ``scale_factor`` / ``add_offset`` must reach the reader.
+
+A netCDF variable may store integer counts alongside the linear transform that
+turns them into physical units.  PACE OCI L2 does exactly this: ``Rrs`` is
+``int16`` with ``scale_factor = 2e-06`` and ``add_offset = 0.05``, so a stored
+-10933 means 0.028134 sr^-1.  Reading the stored value and calling it
+reflectance is wrong by four orders of magnitude, and wrong in a way nothing
+reports.
+
+The fill value is the trap.  It is stored in the same integer space, so scaling
+it along with the data turns -32767 into -0.015534 -- a number that sits inside
+the physical range of the variable and would be indistinguishable from a real
+measurement.  The data-ignore value has to move with the data.
+"""
+import os
+import shutil
+import tempfile
+import unittest
+
+import netCDF4 as nc
+import numpy as np
+
+import tests.context  # noqa: F401 -- adds src/ to sys.path
+
+from wiser.raster.dataset import RasterDataSet
+from wiser.raster.dataset_impl import NetCDF_GDALRasterDataImpl
+
+SCALE = 2.0e-06
+OFFSET = 0.05
+FILL = -32767
+
+BANDS, ROWS, COLS = 3, 4, 5
+
+
+class TestNetCDFScaling(unittest.TestCase):
+    def setUp(self):
+        self._tmp = tempfile.mkdtemp()
+        self.path = os.path.join(self._tmp, "scaled.nc")
+
+        # Stored counts ascending from 0, with one fill pixel in each band.
+        self.raw = np.arange(BANDS * ROWS * COLS, dtype="i2").reshape(BANDS, ROWS, COLS)
+        self.raw[:, 0, 0] = FILL
+
+        with nc.Dataset(self.path, "w") as ds:
+            ds.createDimension("bands", BANDS)
+            ds.createDimension("y", ROWS)
+            ds.createDimension("x", COLS)
+
+            scaled = ds.createVariable("rrs", "i2", ("bands", "y", "x"), fill_value=np.int16(FILL))
+            scaled.scale_factor = SCALE
+            scaled.add_offset = OFFSET
+            # set_auto_scale(False) so netCDF4 stores the counts verbatim rather
+            # than applying the transform on write.
+            scaled.set_auto_scale(False)
+            scaled[:] = self.raw
+
+            plain = ds.createVariable("plain", "f4", ("bands", "y", "x"), fill_value=False)
+            plain[:] = np.full((BANDS, ROWS, COLS), 0.25, dtype="f4")
+
+    def tearDown(self):
+        shutil.rmtree(self._tmp, ignore_errors=True)
+
+    def _open(self, variable):
+        impls = NetCDF_GDALRasterDataImpl.try_load_file(
+            self.path, subdataset_name=variable, interactive=False
+        )
+        return RasterDataSet(impls[0])
+
+    def _expected(self, band):
+        # GDAL presents netCDF rows bottom-up, so the array WISER returns is the
+        # stored band flipped in y.
+        stored = np.flipud(self.raw[band])
+        return stored.astype(np.float32) * np.float32(SCALE) + np.float32(OFFSET)
+
+    def test_scaled_variable_returns_physical_values(self):
+        ds = self._open("rrs")
+        arr = ds.get_band_data(1, filter_data_ignore_value=False)
+
+        self.assertEqual(arr.dtype, np.float32)
+        np.testing.assert_allclose(np.asarray(arr), self._expected(1), rtol=1e-6)
+
+    def test_data_ignore_value_is_reported_in_physical_units(self):
+        ds = self._open("rrs")
+        expected = float(np.float32(np.float32(FILL) * np.float32(SCALE)) + np.float32(OFFSET))
+
+        self.assertAlmostEqual(ds.get_data_ignore_value(), expected, places=9)
+        # The scaled fill lands inside the range of the real data, which is why
+        # leaving it unscaled would let it pass as a measurement.
+        self.assertGreater(ds.get_data_ignore_value(), -1.0)
+
+    def test_fill_pixel_is_masked_and_real_pixels_are_not(self):
+        ds = self._open("rrs")
+        arr = ds.get_band_data(1)
+        mask = np.ma.getmaskarray(arr)
+
+        self.assertTrue(mask[ROWS - 1, 0])
+        self.assertEqual(mask.sum(), 1)
+
+    def test_multiple_band_read_is_scaled(self):
+        # This path also used to call the wrong parent method and raise TypeError.
+        ds = self._open("rrs")
+        cube = ds.get_multiple_band_data([0, 2], filter_data_ignore_value=False)
+
+        self.assertEqual(cube.shape, (2, ROWS, COLS))
+        np.testing.assert_allclose(np.asarray(cube[0]), self._expected(0), rtol=1e-6)
+        np.testing.assert_allclose(np.asarray(cube[1]), self._expected(2), rtol=1e-6)
+
+    def test_unscaled_variable_is_untouched(self):
+        ds = self._open("plain")
+        arr = ds.get_band_data(0, filter_data_ignore_value=False)
+
+        self.assertEqual(arr.dtype, np.float32)
+        np.testing.assert_allclose(np.asarray(arr), 0.25)
+
+
+class TestNetCDFScalingContract(unittest.TestCase):
+    """``_read_scaling`` describes one variable, so its bands must agree."""
+
+    class _Band:
+        def __init__(self, scale, offset):
+            self._scale, self._offset = scale, offset
+
+        def GetScale(self):
+            return self._scale
+
+        def GetOffset(self):
+            return self._offset
+
+    class _Dataset:
+        def __init__(self, bands):
+            self._bands = bands
+            self.RasterCount = len(bands)
+
+        def GetRasterBand(self, number):
+            return self._bands[number - 1]
+
+    def _impl_with(self, bands):
+        impl = NetCDF_GDALRasterDataImpl.__new__(NetCDF_GDALRasterDataImpl)
+        impl.gdal_dataset = self._Dataset(bands)
+        impl._subdataset_name = "rrs"
+        impl._save_state = None
+        return impl
+
+    def test_uniform_scaling_is_returned(self):
+        impl = self._impl_with([self._Band(SCALE, OFFSET)] * 3)
+        self.assertEqual(impl._read_scaling(), (SCALE, OFFSET))
+
+    def test_identity_scaling_reads_as_unscaled(self):
+        impl = self._impl_with([self._Band(1.0, 0.0)] * 3)
+        self.assertIsNone(impl._read_scaling())
+
+    def test_absent_scaling_reads_as_unscaled(self):
+        impl = self._impl_with([self._Band(None, None)] * 3)
+        self.assertIsNone(impl._read_scaling())
+
+    def test_disagreeing_bands_raise(self):
+        impl = self._impl_with([self._Band(SCALE, OFFSET), self._Band(SCALE * 2, OFFSET)])
+        with self.assertRaises(ValueError):
+            impl._read_scaling()
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/src/wiser/raster/dataset_impl.py
+++ b/src/wiser/raster/dataset_impl.py
@@ -1648,6 +1648,71 @@ class NetCDF_GDALRasterDataImpl(GDALRasterDataImpl):
             self._bad_bands = bad_bands
         else:
             self._bad_bands = [1] * self.gdal_dataset.RasterCount
+        self._scaling = self._read_scaling()
+
+    def _read_scaling(self) -> Optional[Tuple[float, float]]:
+        """Return the ``(scale, offset)`` that converts stored values to physical units.
+
+        netCDF ``scale_factor`` and ``add_offset`` are attributes of a variable, so
+        every band of one subdataset carries the same pair even though GDAL exposes
+        them per band.  Returns ``None`` when the variable stores physical values
+        already, which is the common case and leaves the read path untouched.
+
+        Raises
+        ------
+        ValueError
+            If the bands disagree.  A single subdataset is a single variable, so
+            disagreement means the file does not match the netCDF data model, and
+            one physical data-ignore value could not describe it.
+        """
+        pairs = set()
+        for band_number in range(1, self.gdal_dataset.RasterCount + 1):
+            band = self.gdal_dataset.GetRasterBand(band_number)
+            pairs.add((band.GetScale(), band.GetOffset()))
+
+        if len(pairs) > 1:
+            raise ValueError(
+                f"netCDF subdataset {self._subdataset_name!r} reports more than one "
+                f"scale_factor/add_offset pair across its bands: {sorted(pairs)}"
+            )
+
+        scale, offset = pairs.pop()
+        scale = 1.0 if scale is None else float(scale)
+        offset = 0.0 if offset is None else float(offset)
+        if scale == 1.0 and offset == 0.0:
+            return None
+        return scale, offset
+
+    def _unscale(self, arr):
+        """Convert stored values to physical units.
+
+        Returns *arr* unchanged when the variable is unscaled.  Otherwise the
+        result is ``float32``:  the products this applies to store ``int16``
+        counts, and ``float32`` carries the physical value with room to spare
+        while halving the memory a full cube needs.
+        """
+        if self._scaling is None or arr is None:
+            return arr
+
+        scale, offset = self._scaling
+        out = arr.astype(np.float32)
+        out *= np.float32(scale)
+        out += np.float32(offset)
+        return out
+
+    def read_data_ignore_value(self) -> Optional[Number]:
+        """The fill value in the same units as the data :meth:`_unscale` returns.
+
+        The stored fill is scaled by the same arithmetic as the data, so callers
+        comparing against it still match.  Returning the raw fill here would let
+        every fill pixel through as an ordinary physical value.
+        """
+        raw = super().read_data_ignore_value()
+        if raw is None or self._scaling is None:
+            return raw
+
+        scale, offset = self._scaling
+        return float(np.float32(np.float32(raw) * np.float32(scale)) + np.float32(offset))
 
     @contextmanager
     def _quiet_gdal_warnings(self):
@@ -1670,7 +1735,7 @@ class NetCDF_GDALRasterDataImpl(GDALRasterDataImpl):
         """
         with self._quiet_gdal_warnings():
             arr = super().get_image_data()
-        return arr
+        return self._unscale(arr)
 
     def get_image_data_subset(
         self,
@@ -1692,7 +1757,7 @@ class NetCDF_GDALRasterDataImpl(GDALRasterDataImpl):
         """
         with self._quiet_gdal_warnings():
             arr = super().get_image_data_subset(x, y, band, dx, dy, dband, filter_data_ignore_value)
-        return arr
+        return self._unscale(arr)
 
     def get_band_data(self, band_index, filter_data_ignore_value=True):
         """
@@ -1712,7 +1777,7 @@ class NetCDF_GDALRasterDataImpl(GDALRasterDataImpl):
         """
         with self._quiet_gdal_warnings():
             arr = super().get_band_data(band_index, filter_data_ignore_value)
-        return arr
+        return self._unscale(arr)
 
     def sample_band_data(self, band_index, sample_factor: int):
         """
@@ -1724,7 +1789,7 @@ class NetCDF_GDALRasterDataImpl(GDALRasterDataImpl):
         """
         with self._quiet_gdal_warnings():
             arr = super().sample_band_data(band_index, sample_factor)
-        return arr
+        return self._unscale(arr)
 
     def get_all_bands_at(self, x, y):
         """
@@ -1736,7 +1801,7 @@ class NetCDF_GDALRasterDataImpl(GDALRasterDataImpl):
         """
         with self._quiet_gdal_warnings():
             arr = super().get_all_bands_at(x, y)
-        return arr
+        return self._unscale(arr)
 
     def get_multiple_band_data(self, band_list_orig: List[int]) -> np.ndarray:
         """
@@ -1746,8 +1811,8 @@ class NetCDF_GDALRasterDataImpl(GDALRasterDataImpl):
         are extraneous
         """
         with self._quiet_gdal_warnings():
-            arr = super().get_all_bands_at(band_list_orig)
-        return arr
+            arr = super().get_multiple_band_data(band_list_orig)
+        return self._unscale(arr)
 
     def get_all_bands_at_rect(self, x: int, y: int, dx: int, dy: int):
         """
@@ -1759,7 +1824,7 @@ class NetCDF_GDALRasterDataImpl(GDALRasterDataImpl):
         """
         with self._quiet_gdal_warnings():
             arr = super().get_all_bands_at_rect(x, y, dx, dy)
-        return arr
+        return self._unscale(arr)
 
     def get_filepaths(self):
         """


src/wiser/raster/dataset_impl.py
```diff
@@ -1648,6 +1648,71 @@ def __init__(
...
+            )
+
+        scale, offset = pairs.pop()
+        scale = 1.0 if scale is None else float(scale)
+        offset = 0.0 if offset is None else float(offset)
+        if scale == 1.0 and offset == 0.0:
+            return None
+        return scale, offset
+
+    def _unscale(self, arr):
```
mmaclay: 🤖 *Comment generated by Claude and reviewed by Matt*

Reads now return `float32`, but `get_elem_type()` is not overridden, so the dataset still reports `int16`. Every write path asks for that type and casts the read array to it: `save_dataset_as` (2129, then `np.asarray(dst_data, dtype=band_write_dtype)` at 2231), `mosaic_materialize.py:164`/`190`, and `similarity_transform_dialog.py:81` feeding `get_multiple_band_data`.

So opening PACE `Rrs` and using Save dataset as (ENVI) writes a cube of zeros — every physical value is under 0.5 and truncates. `np.asarray(np.float32([-0.0100, 0.0285]), dtype=np.int16)` gives `[0 0]`. Line 2305 then records `data ignore value = -0.015534` in the header beside those zeros. Before this change the path wrote raw counts: wrong units, but recoverable.

Smallest fix is to make the declared dtype match what `_unscale` returns:

```python
def get_elem_type(self) -> np.dtype:
    if self._scaling is None:
        return super().get_elem_type()
    return np.dtype(np.float32)
```

Blocking.

src/wiser/raster/dataset_impl.py
```diff
@@ -1648,6 +1648,71 @@ def __init__(
...
+        The stored fill is scaled by the same arithmetic as the data, so callers
+        comparing against it still match.  Returning the raw fill here would let
+        every fill pixel through as an ordinary physical value.
+        """
+        raw = super().read_data_ignore_value()
+        if raw is None or self._scaling is None:
+            return raw
+
+        scale, offset = self._scaling
+        return float(np.float32(np.float32(raw) * np.float32(scale)) + np.float32(offset))
```
mmaclay: 🤖 *Comment generated by Claude and reviewed by Matt*

Moving the sentinel into physical units is right for everything that reads through this class, but two modules read `impl.gdal_dataset` directly, and those still see raw counts while now getting the scaled sentinel.

`georef_warp.py:275` — `isinstance(target_dataset_impl, GDALRasterDataImpl)` is true here, so it takes `gdal_dataset` (raw `int16`) and passes `noData=target_dataset.get_data_ignore_value()`, now `-0.015534`, to `gdal.Translate` on an Int16 VRT. GDAL rounds that to `0`, so genuine zero-count pixels get dropped as fill while the real fill (-32767) is treated as valid and resampled into its neighbours. `similarity_transform_dialog.py:198` has the same shape. Both were self-consistent before this change.

Worth deciding whether the transform belongs in the dataset rather than in the read wrappers — building the subdataset as a VRT with `gdal.Translate(..., format="VRT", unscale=True, outputType=gdal.GDT_Float32, noData=scaled_fill)` in `_auto_open_elevation` makes `gdal_dataset` physical for all seven read methods, `get_elem_type`, `GetNoDataValue` and every direct consumer at once, and both `_unscale` and this override go away. Failing that, scaled impls need excluding from the direct-`gdal_dataset` branch in both places.

Blocking either way.

src/wiser/raster/dataset_impl.py
```diff
@@ -1648,6 +1648,71 @@ def __init__(
...
+                f"netCDF subdataset {self._subdataset_name!r} reports more than one "
+                f"scale_factor/add_offset pair across its bands: {sorted(pairs)}"
+            )
+
+        scale, offset = pairs.pop()
+        scale = 1.0 if scale is None else float(scale)
+        offset = 0.0 if offset is None else float(offset)
+        if scale == 1.0 and offset == 0.0:
+            return None
+        return scale, offset
```
mmaclay, line 1716: 🤖 *Comment generated by Claude and reviewed by Matt*

The masking in `dataset.py` (985, 1048, 1153) is `np.ma.masked_values`, which is `isclose(rtol=1e-5, atol=1e-8)` rather than equality. Once the sentinel is a physical value, whether a neighbouring count falls inside that tolerance depends on `offset/scale`. PACE is clear by about 12x — tolerance band 1.65e-7 against a count spacing of 2e-6 — but it inverts once `|offset/scale|` passes ~1e5. With `scale=1e-4, offset=100` and a -32767 fill the band is 9.7e-4 against a spacing of 1e-4, so roughly nine counts either side of the fill would mask as fill.

Nothing we read today hits it, and packing with a large offset is common enough (brightness temperature, radiance) that failing at open looks better than dropping pixels quietly later:

```python
if abs(scale) <= 1e-8 + 1e-5 * abs(scaled_fill):
    raise ValueError(...)
```

Non-blocking.

src/wiser/raster/dataset_impl.py
```diff
@@ -1648,6 +1648,71 @@ def __init__(
...
+        Returns *arr* unchanged when the variable is unscaled.  Otherwise the
+        result is ``float32``:  the products this applies to store ``int16``
+        counts, and ``float32`` carries the physical value with room to spare
+        while halving the memory a full cube needs.
+        """
+        if self._scaling is None or arr is None:
+            return arr
+
+        scale, offset = self._scaling
+        out = arr.astype(np.float32)
```
mmaclay: 🤖 *Comment generated by Claude and reviewed by Matt*

The docstring's reasoning holds for `int16`, but nothing enforces that the stored type is `int16`. An `int32`-packed variable is narrowed here without complaint, and counts above 2^24 stop being exactly representable in `float32`; a `float64` variable carrying scale/offset loses precision the same way. Either take the width from `self.gdal_data_type`, or check it in `_read_scaling` and raise. Non-blocking.

src/tests/test_netcdf_scaling.py
```diff
@@ -0,0 +1,163 @@
...
+    def test_multiple_band_read_is_scaled(self):
+        # This path also used to call the wrong parent method and raise TypeError.
+        ds = self._open("rrs")
+        cube = ds.get_multiple_band_data([0, 2], filter_data_ignore_value=False)
+
+        self.assertEqual(cube.shape, (2, ROWS, COLS))
+        np.testing.assert_allclose(np.asarray(cube[0]), self._expected(0), rtol=1e-6)
+        np.testing.assert_allclose(np.asarray(cube[1]), self._expected(2), rtol=1e-6)
+
+    def test_unscaled_variable_is_untouched(self):
```
mmaclay, line 183: 🤖 *Comment generated by Claude and reviewed by Matt*

Two of the seven wrapped read paths are covered here — `get_band_data` and `get_multiple_band_data`. `get_image_data`, `get_image_data_subset`, `sample_band_data`, `get_all_bands_at` and `get_all_bands_at_rect` have no scaling test. `get_all_bands_at` is the one to add first: it's the spectral-profile path, and it's what the PACE lab figures in #777 actually exercise.

Nothing asserts `get_elem_type()` either, or a save-and-reload round trip — a round trip is what would have caught the ENVI truncation.

src/tests/test_netcdf_scaling.py
```diff
@@ -0,0 +1,163 @@
...
+        ds = self._open("plain")
+        arr = ds.get_band_data(0, filter_data_ignore_value=False)
+
+        self.assertEqual(arr.dtype, np.float32)
+        np.testing.assert_allclose(np.asarray(arr), 0.25)
+
+
+class TestNetCDFScalingContract(unittest.TestCase):
```
mmaclay, line 274: 🤖 *Comment generated by Claude and reviewed by Matt*

Small ones, non-blocking: building the impl with `__new__` plus three hand-set attributes will go quietly wrong if `_read_scaling` ever reads a fourth attribute; `_unscale` has no type hints while `_read_scaling` and `read_data_ignore_value` both do; and `read_data_ignore_value`'s docstring doesn't name the units it returns.

mmaclay (commented):
🤖 *Comment generated by Claude and reviewed by Matt*

Self-review, with an assist from Claude.

The float32 arithmetic in `_unscale` and `read_data_ignore_value` is deliberately ordered so the scaled sentinel still compares exactly equal to the scaled fill pixels. That's the non-obvious part, and it's right.

The trouble is that the class now speaks two unit systems. Reads return physical `float32`, but `gdal_dataset` still holds counts, `get_elem_type()` still says `int16` and `GetNoDataValue()` still says -32767, while the cached ignore value is physical. Three write paths cast reads back to the declared dtype and turn a PACE cube into zeros on Save As; two more read `gdal_dataset` directly and get counts with a physical nodata. Notes on both — those are what I'd want settled before merge, and together they argue for baking the transform into a VRT instead of wrapping the readers.

One gap with no line to hang it on: `try_load_file` at `dataset_impl.py:1617` returns a bare `GDALRasterDataImpl` when a file has no subdatasets, which is what GDAL gives for a single-variable netCDF. A packed single-variable file still reads counts, unscaled and unwarned — so the "not PACE-specific" claim in the description doesn't hold yet.

Verification: ruff, DCO and Sphinx green; the `tests` job was still running as I wrote this. Findings come from reading the code plus checking the int16 truncation, the `sorted()` TypeError and the masking tolerance numerically.

src/wiser/raster/dataset_impl.py
```diff
@@ -1648,6 +1648,71 @@ def __init__(
...
+        pairs = set()
+        for band_number in range(1, self.gdal_dataset.RasterCount + 1):
+            band = self.gdal_dataset.GetRasterBand(band_number)
+            pairs.add((band.GetScale(), band.GetOffset()))
+
+        if len(pairs) > 1:
+            raise ValueError(
+                f"netCDF subdataset {self._subdataset_name!r} reports more than one "
+                f"scale_factor/add_offset pair across its bands: {sorted(pairs)}"
+            )
```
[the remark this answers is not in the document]
mmaclay: 🤖 *Comment generated by Claude and reviewed by Matt*

Confirmed — `sorted({(2e-06, 0.05), (None, None)})` raises `TypeError: '<' not supported between instances of 'float' and 'NoneType'`. It still fails loudly, but with the wrong exception and no diagnostic. Reporting the pairs per band in band order avoids the sort and is a more useful message anyway. Worth noting that `test_disagreeing_bands_raise` only covers float-against-float, so this half of the failure path isn't tested.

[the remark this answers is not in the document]
mmaclay: Fixed in 67e46ec. `_read_scaling` now collects into a dict keyed by band number and reports that dict verbatim, so there is no sort to raise `TypeError` and the message names which band carried which pair.

`test_partially_absent_scaling_raises` covers the float-against-`None` half you noted was untested.

src/wiser/raster/dataset_impl.py
mmaclay: Fixed in 67e46ec, as the override you sketched. It delegates to a new `_unscaled_dtype()` rather than hardcoding `float32`, which also answers the `int16`-assumption comment below.

`test_envi_export_round_trips_physical_values` saves and reloads through the ENVI path and asserts the values come back physical — that is the test that would have caught the truncation.

src/wiser/raster/dataset_impl.py
mmaclay: Fixed in 67e46ec via the second option. `reads_are_transformed()` is on the base `RasterDataImpl` returning `False`, overridden on the netCDF impl, and both `georef_warp.py:275` and `similarity_transform_dialog.py:197` now require it false before taking the direct-`gdal_dataset` branch. Scaled impls fall through to the per-band array paths, which read through the scaling.

On the VRT question: it is the better design and I did not take it here, because it breaks project save as written. `dataset.py:1759` takes `recreation_value = impl.gdal_dataset.GetFileList()[0]` for netCDF impls specifically, and against a `/vsimem` VRT that returns a path which does not survive the session. The VRT would also have to outlive the impl across `reopen_dataset()`. Filed as #795 with the full list of what moves.

src/wiser/raster/dataset_impl.py
mmaclay: Taken, in 67e46ec. `_validate_fill_separation()` runs at open and raises with essentially your condition. `_MASK_RTOL` / `_MASK_ATOL` are class attributes mirroring `np.ma.masked_values`' defaults, with a comment saying why they are duplicated here rather than imported.

Two tests: PACE's packing accepted, and `scale=1e-4, offset=100` rejected.

src/wiser/raster/dataset_impl.py
mmaclay: Taken, in 67e46ec — the first option. `_unscaled_dtype()` reads the stored width from `super().get_elem_type()`: integers of 16 bits or fewer go to `float32`, wider integers to `float64` for exactly the 2²⁴ reason, and a stored float keeps its own width rather than being widened for nothing.

Three tests cover the branches: `int16`, `int32`, `float32`.

src/tests/test_netcdf_scaling.py
mmaclay: Fixed in 67e46ec. All seven wrapped read paths now have a scaling test, `get_all_bands_at` included — `test_spectrum_at_pixel_is_scaled` plus `test_spectrum_at_fill_pixel_is_filtered` for the profile path the PACE lab figures exercise.

`get_elem_type()` is asserted in both directions (scaled and unscaled), and `test_envi_export_round_trips_physical_values` is the save-and-reload round trip.

The file went from 9 tests to 26. With `dataset_impl.py` reverted to `main`, 23 of the 26 fail.

src/tests/test_netcdf_scaling.py
mmaclay: Two of three fixed in 67e46ec: `_unscale` is annotated `(Optional[np.ndarray]) -> Optional[np.ndarray]`, and `read_data_ignore_value`'s docstring now says the value comes back in the units `_unscale` returns rather than the stored units.

The `__new__` construction stays. `_impl_with` sets five attributes now, and its docstring names the tradeoff — bypassing `__init__` is what keeps the open-time checks testable without a file on disk, at the cost of a method growing a sixth attribute needing a sixth line here. Happy to build a real fixture file instead if you would rather not carry that.

src/wiser/gui/similarity_transform_dialog.py
```diff
@@ -193,8 +193,13 @@ def create_translated_dataset(
     driver: gdal.Driver = gdal.GetDriverByName("GTiff")
     new_dataset = None
     try:
-        if isinstance(dataset.get_impl(), GDALRasterDataImpl):
-            impl = dataset.get_impl()
+        impl = dataset.get_impl()
+        if isinstance(impl, GDALRasterDataImpl) and not impl.reads_are_transformed():
```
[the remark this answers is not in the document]
mmaclay: Added in bdf4233d: `TestNetCDFScalingConsumers::test_similarity_translation_writes_physical_values`, which runs a packed netCDF through `create_translated_dataset` and asserts the output is `Float32` carrying physical values with a physical nodata. It lives in `test_netcdf_scaling.py` rather than the similarity-transform tests because the packed fixture is the expensive part and the property under test is the scaling contract; the fixture is now a mixin shared by both classes there.

I confirmed it can fail: with the `reads_are_transformed()` guard removed from line 197, this test and the `georef_warp` one both fail, and GDAL emits "Nodata value was not set to output band, as it cannot be represented on its data type" — the physical nodata hitting an Int16 band.

Writing it turned up a separate defect, which I have left alone here: the fallback writes `-9999` for fill pixels while the header declares the dataset's ignore value, because `write_raster_to_dataset` fills masked entries with `DEFAULT_IGNORE_VALUE`. That affects any dataset on this path with a non-`-9999` ignore value, predates this PR, and the rotate-and-scale path 95 lines above already does it correctly. Filed as #797. The test excludes that one pixel with a comment pointing there.

src/wiser/raster/dataset_impl.py
```diff
@@ -1648,6 +1676,152 @@ def __init__(
...
+        to ``float64``, because ``int32`` counts above 2**24 are not exactly
+        representable in ``float32`` and narrowing them would drop precision
+        with nothing reporting it.
+        """
+        stored = super().get_elem_type()
+        if np.issubdtype(stored, np.floating):
+            return stored
+        if stored.itemsize <= 2:
+            return np.dtype(np.float32)
+        return np.dtype(np.float64)
```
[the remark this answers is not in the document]
mmaclay: You are right that the docstring was false, and I took the second of your two options in bdf4233d — a stated precision policy with a regression test — rather than rejecting 64-bit packed integers at open. Three things decided it.

float64 carries every int64 count exactly up to 2\*\*53. A dtype check rejects that entire population, which is realistically all of it: anyone packing into `int64` is taking a default width, not using 9 quadrillion counts.

Above 2\*\*53 there is nothing better to do. At a count of 2\*\*53 with this file's packing the physical value is ~1.8e10, where float64's own spacing is 3.8e-06 — wider than the 2e-06 step between adjacent counts. Those counts are not distinct in physical units under any float, and no wider float is portable, so the merge is what float64 can express rather than something the cast introduces.

That also means the raise could not tell the user how to open the data. The only remedy is `gdal_translate -unscale -ot Float64` outside WISER, which produces byte-identical merged values.

I checked whether we could validate the real condition instead of using dtype as a proxy, and we cannot cheaply: GDAL reports `GetMinimum()`, `GetMaximum()` and `GetStatistics(force=0)` all `None` for netCDF bands, so knowing the count range means reading the whole cube at open.

So `_unscaled_dtype`'s summary no longer claims exactness, and the docstring states where the boundary is and why nothing moves it. `test_packed_64_bit_counts_are_carried_only_below_2_53` pins it in both directions: 2\*\*53−1 is carried, 2\*\*53+1 is not.

Verified against GDAL 3.12, which does expose packed `i8`/`u8` netCDF variables as `Int64`/`UInt64` bands with the scale and offset attached, so the path is reachable rather than theoretical.

src/wiser/raster/georef_warp.py
```diff
@@ -270,8 +270,15 @@ def warp_dataset_to_path(
     progress.report(0, target_dataset.num_bands(), "Warping...")
 
     ratio = MAX_RAM_BYTES / output_bytes
-    if isinstance(target_dataset_impl, GDALRasterDataImpl):
-        # Saving the full gdal dataset
+    if (
+        isinstance(target_dataset_impl, GDALRasterDataImpl)
+        and not target_dataset_impl.reads_are_transformed()
```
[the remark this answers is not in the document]
mmaclay: Added in bdf4233d: `TestNetCDFScalingConsumers::test_warp_writes_physical_values` puts a packed netCDF through `warp_dataset_to_path` with an affine GCP set and asserts the output is `Float32` with a physical nodata and physical values throughout. You are right that `make_numpy_scene` never reaches this dispatch — a numpy scene fails the `isinstance` check and takes a different branch entirely, so the guard was untested in both directions.

Confirmed it fails with the guard removed from line 275, alongside the similarity-transform one.

